PR #4, opened 2018-07-31
Fix wording, and make the language more consistent.

stucchimax:


partim:
Thank you for the PR!

PR #11, opened 2018-10-27
- Fix wrong port number in example

stucchimax:
- While here, fix some language here and there


partim:
The IANA assigned port for RPKI-RTR really is 323 (see section 9 of RFC 8210). Maybe my wording wasn't very clear, but what I meant to say way that this uses 3323 as the port of you don't specify one instead of the RFC-mandated 323 because you need to be root to bind to that port and there is absolutely no reason to run Routinator as root.

partim:
I’ll merge the PR to get all your language fixes and then update the text around port 323/3323 to make more clear what is intended.

Thank you for the PR!

PR #23, opened 2018-11-22
Introduce Dockerfile for docker hub automated container builds

wk:
Introduces a minimalist Dockerfile for use with `hub.docker.com` (assumes NLnetLabs has control over a similarly named organization account on Docker Hub) based on `alpine` and documents its use.

Very lightly tested. Requires enabling and configuring Docker Hub for this repository.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,8 +18,8 @@ COPY --from=build /tmp/routinator/target/x86_64-unknown-linux-musl/release/routi
 # Install rsync as routinator depends on it
 RUN apk add rsync
 
-# Due to ARIN TAL licensing terms, we can't do this here. An individual user, however,
-# might want to anyway after reviewing https://www.arin.net/resources/rpki/tal.html
+# Due to ARIN TAL distribution terms, we can't do this here. An individual user, however,
+# might want to anyway - after reviewing https://www.arin.net/resources/rpki/tal.html
 # If this is in place, mouting a volume at run time with this file is no longer necessary.
 # 
 # ADD https://www.arin.net/resources/rpki/arin-rfc7730.tal /root/.rpki-cache/tals/arin.tal
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,9 @@ RUN apk add rsync
 # 
 # ADD https://www.arin.net/resources/rpki/arin-rfc7730.tal /root/.rpki-cache/tals/arin.tal
 
-VOLUME ["/root/.rpki-cache"]
+# Prepare a directory for TALs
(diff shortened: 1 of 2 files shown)


partim:
Awesome! Thank you very much for the PR.

I’ll look into getting an organization for us going on Docker hub.

One note regarding the paths: I don’t think the local copy of the RPKI repository should go into /etc, I think it shouldn’t even bleed out of the container.

Routinator allows to provide separate paths for the TAL directory and the repository cache via the -t and -c options. So my proposal would be to have /etc/routinator/tals on the host system, mount that as a volume and provide it to Routinator via -t. The cache can go anywhere inside to container, so something like `-c /root/rpki-cache` should do the trick.

wk:
Thank you for reviewing this!

One potential disadvantage to overriding the default location of the TALs (and cache) via a command-line parameter would be that a user which wishes to run the container with alternative command line parameters than the container's `CMD` defaults would need to be aware of the (non-default) locations and pass them at every invocation or face an unexpected error.

A possible alternative to address your specific concern without modifying the defaults would be to create the `/root/.rpki-cache/tals` directory tree at container build, and then leak the volume `/root/.rpki-cache/tals` instead of `/root/.rpki-cache` from the container into the host. The user can then be advised by default to bind the host volume `/etc/routinator/tals` to the container's `/etc/.rpki-cache/tals`. From a user perspective, this should be indistinguishable from your proposal.

A different approach would be to begin considering a `daemon` style packaging, in which case the executable would live in `/usr/bin`, the cache in `/var/routinator/cache` and TALs in `/var/routinator/tals`, which at this stage would be specified via the -c and -t command line parameters of the container; over time, these defaults could also be introduced into cargo's configuration as hints for cargo-deb so that an automated `cargo deb` build of a routinator debian package would also result in a package that utilizes these paths (and include the necessary systemd units for daemon operation).

wk:
The requested adjustments are now in place.

partim:
I guess we overlooked that now the ARIN TAL is the _only_ TAL in the TAL directory. There’s a few options to fix this but maybe allowing multiple TAL directories is the cleanest way of doing this?

wk:
Corrected to gain the originally desired behavior in #25. 

Two directories could be problematic for instances in which a user explicitly wishes to remove one of the core shipped TALs by deleting the file (where from?). 

The current behavior allows both deletion of any of the 5 root TALs seeded on first run (if directory isn't present), addition of new TALs after first run, or provision of a completely custom directory with a set of TALs ahead of first run (as happens here). The existing behavior is likely desirable.

PR #25, opened 2018-11-22
Bring TALs in correctly into Dockerfile

wk:
This corrects an omission in NLnetLabs/routinator#23 in regards to how routinator treats an existing TALs directory with only one TAL present (as a result of following ARIN TAL instructions). 

Instead, the TALs are now copied over from source during container build, and the instructions have been updated in a manner to facilitate overloading the shipped `arin.tal` file with the genuine article downloaded by the user from the ARIN website.

This leads to correct behavior in both absence of a correct `arin.tal` (error is displayed), as well as presence of downloaded file - and still allows the user to add TALs by mounting additional files into the tals directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,8 @@ RUN apk add rsync
 # Prepare a directory for TALs
 RUN mkdir -p /root/.rpki-cache/tals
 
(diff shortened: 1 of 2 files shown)


partim:
Thank you so much for your effort!

PR #32, opened 2018-12-01
Add openbgpd compatible outform

### LINKED ISSUE #31: Extend outform with openbgpd format
It would be cool if routinator's `-f` option allows `openbgpd` format so that users of the software can immediately plug the output into the openbgpd daemon.

The openbgpd format is as following:

```
roa-set {
    57.0.0.0/8 source-as 2647
    100.128.0.0/9 maxlen 24 source-as 21928
    35.0.0.0/10 source-as 237
    78.192.0.0/10 maxlen 11 source-as 12322
    79.192.0.0/10 source-as 3320
    8.128.0.0/10 source-as 37963
    ... etc ...
}
```

What is of note is that the `maxlen` argument is optional if the prefixlen and the MaxLength are the same value.

job:
This fixes #31

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -213,6 +213,7 @@ fn output(
     match config.outform {
         OutputFormat::Csv => output_csv(roas, &mut output),
         OutputFormat::Json => output_json(roas, &mut output),
+        OutputFormat::Openbgpd => output_openbgpd(roas, &mut output),
         OutputFormat::Rpsl => output_rpsl(roas, &mut output),
         OutputFormat::None => { Ok(()) }
     }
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -261,6 +262,37 @@ fn output_json<W: io::Write>(
...
+        if addr.address_length() < addr.max_length() {
+            write!(output, " maxlen {}",
+                addr.max_length(),
+            )?;
+        }
+        write!(output, " source-as {}",
+            &addr.as_id().to_string()[2..], // sorry, this feels ugly
+        )?;
+    }
+    writeln!(output, "\n}}")?;
```
partim: If you move the `\n` into the loop, you can drop the entire `first` thing.

src/main.rs
```diff
@@ -261,6 +262,37 @@ fn output_json<W: io::Write>(
...
+        write!(output, "    {}/{}",
+            addr.address(), addr.address_length(),
+        )?;
+        if addr.address_length() < addr.max_length() {
+            write!(output, " maxlen {}",
+                addr.max_length(),
+            )?;
+        }
+        write!(output, " source-as {}",
+            &addr.as_id().to_string()[2..], // sorry, this feels ugly
```
partim: I _think_ this should work instead without double string conversion:

```suggestion
            u32::from(addr.as_id())
```

PR #33, opened 2018-12-05
Remove a redundant definition of --strict.

stucchimax:


partim:
Thanks for the PR!

Just a quick heads-up: The entire command line is changing with branch `new-structure`, which will cause some heavy re-writing of both the manpage and the README. Just so you don't spend time in reviewing right now …

PR #61, opened 2019-02-03
bump rust to 1.32 and alpine to 3.9 for dockerfile freshness

wk:


wk:
Didn't realize how much of a difference that makes! Last build on 1.30.1:
`-rwxr-xr-x    1 root     root      10036688 Feb  1 12:54 /usr/local/bin/routinator`
Test build on 1.32.0:
`-rwxr-xr-x    1 root     root       7119488 Feb  3 11:57 /usr/local/bin/routinator`

Very cool! Thanks for the heads up!

PR #65, opened 2019-02-05
lint with clippy during travis

wk:
As one of the emergent implicit goals of this project turned out to be the introduction of rust-lang within the network engineering community, make use of the [rust-clippy](https://github.com/rust-lang/rust-clippy) linter to encourage rust best practices and code correctness.

Initially configured to only produce linter recommendations as part of travis build output, can be modified to fail the build when linter warnings appear after initial linting work is complete to encourage code correctness in PRs and keep contributors honest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,9 @@ rust:
   - stable
   - beta
   - 1.30.0  # The minimal Rust version we claim we need.
-
+before_script:
+  # Get clippy-preview if clippy is not available for 1.30.0 not to break
+  - rustup component add clippy || rustup component add clippy-preview
+script:
+  - cargo clippy -- -D warnings
+  - cargo test --verbose
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,7 +2,14 @@
 //!
 //! This crate contains all the moving parts of the Routinator. The
 //! application itself, via `main.rs` is only a very tiny frontend.
-#![allow(unknown_lints)] // Clippy for 1.30.
+
+// Clippy due to multi-versioning
+#![allow(renamed_and_removed_lints)]
+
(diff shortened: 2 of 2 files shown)


wk:
Demonstration of `rust-clippy` output can be seen in the Travis CI details of this PR's build; unavailability of the `clippy` component under rust 1.30.0 (causing the build to fail) requires further investigation.

wk:
A workaround for builds failing under 1.30.0 is now included.

partim:
I suggest we leave this one open until we can safely add the `-D warnings` argument. Shouldn’t be too long.

partim:
Clippy now seems happy (or has been silenced). Could you perchance add a commit with the `-D warnings` option enabled?

partim:
Maybe we should just bite the bullet and go to Rust 1.32 as minimum version.

wk:
stable and beta currently succeed, 1.30 has a handful of `needless_pass_by_value` and `map_clone` hits which would need to be suppressed, either globally (with `-A`), code-wide, or per-instance.

It seems clippy is somewhat more fickle and rapidly moving than I've anticipated when initially suggesting this, but so is the entire language...

This *could* build cleanly with [these](https://travis-ci.com/NLnetLabs/routinator/jobs/177237106) suppressions in place under 1.30.

wk:
I've suppressed `needless_pass_by_value` and `map_clone` globally, as well as `renamed_and_removed_lints`, but enabled job failing on clippy warnings; which should be the desired outcome. All three builds now pass.

Global rather than topical suppression will eliminate need of code-base wide clean up of clippy hints when 1.30 is eventually abandoned.

partim:
Many thanks for figuring this out!

PR #92, opened 2019-03-15
document prometheus monitoring better

### LINKED ISSUE #91: Default prometheus exporter port for routinator
Prometheus maintains a list of allocations of [default ports](https://github.com/prometheus/prometheus/wiki/Default-port-allocations) for exporters, and routinator has been added to the list as port `9556`. 

This could be reflected as a recommendation or a default in `config.rs`, or in documentation (`README.md`, `doc/routinator.1`) as a value for `listen-http`.

wk:
Documents the allocated default prometheus monitoring port, and introduces listen-http to the example `routinator.conf` configuration file.

Closes #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ Status](https://ci.appveyor.com/api/projects/status/github/NLnetLabs/routinator?
 Introducing ‘Routinator 3000,’ RPKI relying party software written in Rust.
 If you have any feedback, we would love to hear from you. Don’t hesitate to
 [create an issue on Github](https://github.com/NLnetLabs/routinator/issues/new)
-or post a message on our [RPKI mailing list](https://nlnetlabs.nl/mailman/listinfo/rpki). 
-You can lean more about Routinator and RPKI technology by reading our documentation on 
+or post a message on our [RPKI mailing list](https://nlnetlabs.nl/mailman/listinfo/rpki).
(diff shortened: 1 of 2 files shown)


partim:
The idea for the sample config was that the commented-out values show the default value for that particular item (hence the `...` in some of them). Not sure if that is reasonable, but that’s what I normally assume when seeing a sample config.

wk:
That's reasonable, but in this case it complicates the situation somewhat requiring the introduction of a new use case into `config.rs` to handle a knob for which there should (could?) be a default value, but which itself shouldn't be (isn't?) enabled by default.

* Example in routinator.conf can be change to '...' documenting allocated port in a comment instead.
* listen-http on 127.0.0.1:9556 can default on in 'rtrd' mode.
* listen-http on 127.0.0.1:9556 can be default if --listen-http is not supplied with parameters, introducing new parameter use case.

What would be your preference?

wk:
Adapted #92 for the first option (as minimally disruptive) so that this can move ahead while the underlying question about the future of listen-http is contemplated in due time.

partim:
I am contemplating providing a way to disable RTR, perhaps `--listen-tcp=none`, since the HTTP-only use case exists. This would also provide a way forward for HTTP: enable it by default and allow turning it off.

PR #123, opened 2019-05-22
make arin.tal more instructive to match up with @rfc1036 routinator debian packaging

wk:
Discussed with @rfc1036 out-of-band, and meets both their stated requirements (to provide a copy-pasteable command that ensures user success) in the Debian package [here](https://salsa.debian.org/md/routinator/tree/master/debian). This eliminates the need to patch the file during packaging while being generic enough for all routinator users.

### DIFF
--- a/tals/arin.tal
+++ b/tals/arin.tal
@@ -4,7 +4,13 @@ This file isn’t ARIN’s TAL but rather a placeholder. Until you replace
 it with the actual TAL you can find at the above URL, Routinator will
 not run.
 
-Please download the TAL in RFC 7730 format and copy it here.
+Please download the TAL in RFC 7730 format after reviewing the usage agreement
+at https://www.arin.net/resources/manage/rpki/tal/ and copy it here.
+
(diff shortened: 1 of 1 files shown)


tals/arin.tal
```diff
@@ -4,7 +4,13 @@ This file isn’t ARIN’s TAL but rather a placeholder. Until you replace
 it with the actual TAL you can find at the above URL, Routinator will
 not run.
 
-Please download the TAL in RFC 7730 format and copy it here.
+Please download the TAL in RFC 7730 format after reviewing the usage agreement
+at https://www.arin.net/resources/manage/rpki/tal/ and copy it here.
+
+Example (assumes system-wide routinator installation):
```
partim: I’d like to suggest a slightly more explicit wording:

```
For a system that stores the TALs in /etc/routinator/tals, this can be
achieved by running:
```

mahtin:
+1

wk:
This looks fine to me, and should still meet both stated goals.

PR #124, opened 2019-05-23
items are not mutable

wk:
Unnecessary (unused) mutability is a warning in Rust 1.36, this clears it up.


partim:
Thanks! As always, you are about two version ahead of me …

PR #126, opened 2019-05-23
Add systemd unit file

partim:


Cargo.toml
```diff
@@ -59,6 +59,7 @@ assets = [
     ["README.md", "usr/share/doc/routinator/", "644"],
     ["doc/misc.md", "usr/share/doc/routinator/misc.md", "644"],
     ["doc/routinator.1", "usr/share/man/man1/routinator.1", "644"],
+    ["etc/routinator.service", "lib/systemd/system/routinator.service", "644"]
```
CameronNemo: Is this configurable? Either to not install it or install it into `/usr/lib`. Maybe better to simply place the unit in a `contrib` directory and let the distributor decide whether or not to install it, and to where.

partim: This is only for `cargo deb`. Having the file won’t work for the RTR daemon, since you also will need a config file and the TALs. So maybe we should indeed take it out here and only have the `cargo deb`-built package work for command-line mode?

PR #127, opened 2019-05-27
Accept RTR listener from systemd.

### LINKED ISSUE #114: Manage RTR sockets with rust-listenfd
This will allow Routinator’s RTR daemon to be run on port 323 without being root.

partim:
Fixes #114.


rfc1036:
I strongly recommend against implementing a --listen-systemd option: no damon with socket activation support ever does this. The correct procedure is to check if any socket-activated FDs have been provided and use them if have been, or else just continue with the usual workflow if they have not.
And if routinator supports binding on multiple sockets by itself then it should support receiving multiple sockets from systemd as well.

For more information: http://0pointer.de/blog/projects/socket-activation.html

partim:
I checked with the Unbound people before implementing and they do actually have an argument – well, a config file option – for this purpose.

Given that socket activation uses environment variables and that I don’t want to use features if I can avoid it, I think the risk that socket activation is invoked accidentally leading to Routinator to fail to start (not yet, but a future version will move socket creation forward and refuse to start if it can’t create all sockets requested) is too high. Also, the reverse is true: if you specify `--listen-systemd` and there is no TCP listener passed in, it can fail with a meaningful message.

rfc1036:
I cannot speak for the unbound developers, but it is the first daemon that I have ever seen which does this. There is no risk that socket activation can be activated by mistake since it depends on some very specific environment variables which are never found anywhere else. I am one of the Debian systemd package maintainers and I am quite sure that I would have noticed bug reports about daemons broken by leaked socket activation variables.

Anyway, the bigger issue is that from what I understand of this commit it supports only one socket-activated FD while the daemon supports listening to multiple sockets. This is critical to support IPv6 without resorting to v6-mapped ::ffff:0.0.0.0 addresses.

partim:
I’m still not convinced I want to even run this code on systems that don’t have systemd.

For the second part, fair point. I will adjust the code to use all TCP listeners received via socket activation for RTR. If we’d ever want to extend socket activation to the HTTP listeners, then we’ll have to reconsider.

rfc1036:
If in the future you will want to support socket activation also for HTTP then you just need to use the Rust equivalent of sd_listen_fds_with_names(3).

See FileDescriptorName in systemd.socket(5) for details.

partim:
Yes. That isn’t currently implemented in the Rust crate, though, so it has to wait.

PR #137, opened 2019-05-28
Allow filtering for ASN and prefix in vrps and HTTP output.

### LINKED ISSUE #132: Allow filtering in VRP output.
In both the `vrps` command and the HTTP server’s VRP output endpoints, allow limiting output by specifying a list of prefixes and ASNs. For `vrps`, this should be via the command line. For HTTP, use the query string.

partim:
Fixes #132.

### DIFF
--- a/doc/routinator.1
+++ b/doc/routinator.1
@@ -50,8 +50,11 @@ command
 .IR output-file ]
 .RB [ \-f
 .IR format ]
-.RB [ \-n
-]
+.RB [ \-n ]
+.RB [ \-a
+.IR asn ]
+.RB [ \-p
+.IR prefix ]
 .PP
 .B routinator
 [options]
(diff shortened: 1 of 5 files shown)


doc/routinator.1
```diff
@@ -661,6 +684,19 @@ output format.
 .B /version
 Returns the version of the Routinator instance.
 
+.P
+The paths that output the current set of VRPs accept filter expressions to
+limit the VRPs returned in the form of a query string. The field
+.B filter-asn
+can be used to filter for ASNs and the field
+.B filer-prefix
```
AlexanderBand: filter-prefix, not filer

doc/routinator.1
```diff
@@ -330,6 +333,26 @@ This format produces no output whatsoever.
...
+and not just the number
+e.g., the correct form is
+.BR "AS65555" .
+.TP 
+.BI \-p \ prefix\fR,\  \-\-filter\-prefix= prefix
+Only output VRPs with an address prefix that covers the given prefix, i.e.,
+whose prefix is equal or less specific than the given prefix. This will
+include VRPs irregardless of their ASN and max length. In other words, the
+output will include all VRPs that need to be considered when deciding whether
+an announcement for the prefix.
```
AlexanderBand: This sentence feels unfinished.

doc/routinator.1
```diff
@@ -330,6 +333,26 @@ This format produces no output whatsoever.
...
+as the AS number prefixed with 
+.B AS 
+and not just the number
+e.g., the correct form is
+.BR "AS65555" .
+.TP 
+.BI \-p \ prefix\fR,\  \-\-filter\-prefix= prefix
+Only output VRPs with an address prefix that covers the given prefix, i.e.,
+whose prefix is equal or less specific than the given prefix. This will
+include VRPs irregardless of their ASN and max length. In other words, the
```
AlexanderBand: I was taught "irregardless" is not a 'proper' word, and that "regardless" works just as well.

CameronNemo: It is not a word at all, and the etymologically extrapolated meaning is the opposite of how it is typically used (including in this instance).

doc/routinator.1
partim: Fixed by d628aeb.

doc/routinator.1
partim: Fixed by d628aeb.

doc/routinator.1
partim: You are, of course, right.

Fixed by d628aeb.

Issue #82, opened 2019-03-05
Add systemd unit file for rtrd to Debian packaging.


partim:
The Debian packaging currently doesn't include a init script or systemd unit file for starting Routinator in rtrd mode. We need to explore how these files differ on various Debian-based distributions and whether we can add mechanism to include appropriate ones, at least for a subset of common distributions.

CameronNemo:
As long as the systemd unit is generic you can use the same one across different Debian and Ubuntu versions. Ubuntu and Debian publish manpages where you check which configuration directives are supported.

rfc1036:
A systemd unit is now available from https://salsa.debian.org/md/routinator/blob/master/debian/routinator.service .
systemd units are intended to be cross-platform (also for RPM-based distributions). Since it uses feature implemented in recent releases of systemd some lines will have to be removed when using it on older distributions.

When people will have a bit more operational experience using it then I think that it should be moved to the main routinator tree.

partim:
That’s looks a lot more complete than what I cobbled together! Thank you!

I’m happy to add it now – that would also give more exposure to it so more people test it. If you agree, would you like to make a PR so it gets proper attribution or want me to do that?

wk:
For this systemd unit to work out of the box, some assumptions have to be met regarding paths for the various components (configuration, roas, ...); this means different defaults will be required at least in the shipped .conf file, and possibly in the daemon itself.

wk:
As defaults, the packaging paths in @rfc1036's [repository](https://salsa.debian.org/md/routinator/tree/master/debian) seem sensible, specifically, these include:

`repository-dir` defaulting to `/var/lib/routinator/repository/`
`tal-dir` defaulting to `/etc/routinator/tals/`

This also implies (although does not strictly require) that the default config file, if present, to be `/etc/routinator/routinator.conf`.

wk:
A variation on [this](https://salsa.debian.org/md/routinator/blob/master/debian/patches/arin-download) patch that's used as part of the Debian package and which should be acceptable for general use has been PR'ed as #123

XioNoX:
FYI, on my test Debian 9.8 system, the line `SystemCallFilter=@system-service`
causes Routinator to segfault at start time:
```
May 24 02:57:58 traffic-rpki systemd[1]: Started Routinator 3000.
[...]

Issue #18, opened 2018-11-09
Improve daemon mode for system service


partim:
Daemon mode currently only detaches from the terminal. It needs to do all the things that daemon mode should do.

Hint: this basically includes all the options provided by `daemonize` crate we are already using. All of it needs to be configurable, of course.

PR #139, opened 2019-05-29
Update dockerfile

partim:
As of now, this only updates the options of the CMD. However, we need to figure out a good way to deal with the modified TAL initialization mechanism in a way that fulfills ARIN’s requirements.

I guess running docker with an alternative command would be a good way?

Cc: @wk

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,7 +47,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "backtrace"
-version = "0.3.22"
+version = "0.3.26"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "autocfg 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


wk:
Today, the Dockerfile does not rely on routinator's ability to produce the TALs; it instead copies them from the source tree directly to where they are expected during build (and the runtime instructions advise overwriting the ARIN dummy by mounting one externally as a volume). 

This means the presence / absence of the 'init' command does not directly influence the setup, and in case the TALs shipped with routinators will change, they will also change in the source tree, and thus will be picked up by the automated Dockerhub build.

If there is desire to transition this, the init command can be ran during build as a last step, or documentation can be written to encourage doing so upon first run as alternative command.

This ties into #18 and #82, as the Dockerfile is opinionated toward operation as a daemon/service, converging the way across systemd unit defaults, routinator default configuration loading (do we look at /etc/routinator/routinator.conf by default if it's present without it being specified), and default paths for TALs and ROAs will make the correct thing to do here increasingly clear.

partim:
I should have looked at the Dockerfile more closely.

This is now problematic, as it copies and uses the ARIN TAL without getting user agreement to the RPA. I’m fine with people doing this in their own Docker builds, but the official one can’t really just install the ARIN TAL and move on. With the previous mangled ARIN TAL, the build would just not start unless you did something.

Which why I thought not installing the TALs as part of the Dockerfile but asking people to run something like

```
docker run --name routinator init --accept-arin-rpa
```

would be a more appropriate approach.

partim:
I think I like option 2 best.

If we set
```
ENTRYPOINT ["routinator"]
CMD ["server", "--rtr", "0.0.0.0:3323"]
```
we could instruct people to first run the image once with `--rm` and `init -f --accept-arin-rpa` to push the TALs into the mounted volume and then just run it again without arguments.

We can even leave hints in the Dockerfile on how to make your own image without the need for any of these shenanigans.

wk:
Elegant. I like it.

PR #145, opened 2019-06-02
use explicit dyn in BoxedError to appease compiler

wk:
trait objects without an explicit dyn have been deprecated in 1.37


partim:
Thanks! We should probably add nightly to the Travis build matrix. Except, it already takes way too long.

PR #151, opened 2019-06-03
enable nightly for CI/CD and speedups

wk:
add rust nightly to CI/CD builds while ignoring any errors, and speed up the build with cargo caching

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,15 @@ os:
 rust:
   - stable
   - beta
+  - nightly
   - 1.34.0  # The minimal Rust version we claim we need.
+matrix:
+  allow_failures:
+    - rust: nightly
+  fast_finish: true
 before_script:
   - rustup component add clippy
 script:
   - cargo clippy -- -D warnings
   - cargo test --verbose
+cache: cargo


wk:
The current version will fail nightly if clippy is not available, without attempting a build. Since the rust tools [occasionally](https://rust-lang.github.io/rustup-components-history/) fall out of sync in nightly, it might be better to either not run clippy on nightly at all, or attempt to run clippy on nightly as a best effort, via something like this (unvalidated):

```yaml
language: rust
os:
  - linux
  - windows
rust:
  - stable
  - beta
  - nightly
  - 1.34.0  # The minimal Rust version we claim we need.
matrix:
  allow_failures:
    - rust: nightly
  fast_finish: true
before_script:
  - bash -c 'if [[ "$TRAVIS_RUST_VERSION" != "nightly" ]]; then
      rustup component add clippy;
    else
      rustup component add clippy || true;
    fi'
script:
  - bash -c 'if [[ "$TRAVIS_RUST_VERSION != "nightly" ]]; then
      cargo clippy -- -D warnings;
    else
      cargo clippy -- -D warnings || true;
    fi'
  - cargo test --verbose
cache: cargo
```

PR #159, opened 2019-06-06
Fix example config file parameters

devicenull:
The listen-http name was backwards, and listen-tcp seems to have moved to rtr-listen


partim:
Thanks for the PR. I forgot about the example config when making all those changes.

Issue #160, opened 2019-06-07
Prometheus # TYPE for some metrics wrong


momorientes:
The following metrics show `# TYPE gauge` instead of `# TYPE <metric_name> gauge` .
```
last_update_start
last_update_duration
last_update_done
serial
```

I will provide a PR for this.

qybl:
FYI #161 actually fixed #154. #156 did not :)

partim:
Might be a version thing. #156 made my Prometheus installation go green again.

qybl:
Hum, I tried with a current master and `cargo build` and `cargo build --release`. Prometheus was 2.10.0 fresh from prometheus.io. Only after I pulled again with #161 merged did prometheus stop erroring out on routinators /metrics.
As I'm new to building rust software might have been some error on my side, I don't know.

partim:
Nono, this is my laziness. I was testing with Prometheus 1.5.2 (which seems to be the version that Debian Stretch has). Should have also tested with latest Prometheus.

Issue #157, opened 2019-06-06
Prefix Prometheus metrics with `routinator_`.


partim:

partim:
This is going to be a breaking change, so this is for 0.5.

AlexanderBand:
As per https://phabricator.wikimedia.org/T225108#5237480, via @XioNoX

wk:
This would also be consistent with the prometheus [best practices](https://prometheus.io/docs/practices/naming/) guide.

partim:
Fixed by #162.

PR #162, opened 2019-06-07
prefix prometheus metrics with routinator_

momorientes:
implements https://github.com/NLnetLabs/routinator/issues/157
fixes https://github.com/NLnetLabs/routinator/issues/160


partim:
Since changing the names of the metrics is a breaking change, I want to hold off on this PR until we have decided whether we’re going to do a 0.4.1 before 0.5.

momorientes:
fine by me, I was just fixing #160 so I thought I'd just fix #157 while I'm at it.
Thanks for the quick merge

PR #196, opened 2019-08-16
Switch Docker build image to Alpine.

partim:
This seems to work, but I am not sure this is the right thing to do?

Cc: @wk

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,21 @@
 # -- stage 1: build static routinator with musl libc for alpine
-FROM rust:1.36.0-stretch as build
+FROM alpine:3.10.1 as build
 
-RUN apt-get -yq update && \
-    apt-get -yq install musl-tools
-
-RUN rustup target add x86_64-unknown-linux-musl
+RUN apk add rust cargo openssl-dev
 
 WORKDIR /tmp/routinator
 COPY . .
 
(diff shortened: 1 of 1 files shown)


wk:
This approach results in a dynamic routinator linked against both musl, libgcc, and libssl rather than a static routinator, this is due to several factors:
* OpenSSL requires additional environment variables to actually perform a static build, including `OPENSSL_LIB_DIR` and `OPENSSL_INCLUDE_DIR`, without them, a dynamic build is performed.
* The rustc linker needs additional hints to correctly statically link the end result into the binary
* The dynamic OpenSSL build for reqwest will require libgcc to be present as well.

This example is provided for reference only; there's nothing wrong with the dynamic approach as libssl ships with alpine out of the box, and the additional libgcc dependency is fairly trivial.

wk:
The right thing to do here would probably be to abandon the static build, and instead do a dynamic build targetting alpine (as in example in #198) for dynamic musl, together with dynamic libssl (as it comes out of the box), and indeed adding libgcc.

Basically, this PR without the `ENV OPENSSL_STATIC=1 PKG_CONFIG_ALLOW_CROSS=1` line, and with `target=x86_64-unknown-linux-musl` replaced by `x86_64-alpine-linux-musl`.

partim:
Thanks, @wk! This PR is now superseded by #199.

PR #198, opened 2019-08-16
build dynamic musl routinator with static libssl for Docker

wk:
Superset of #196. While #196 builds a dynamic routinator with dynamic libssl that also has a libgcc dependency, resulting in requiring both musl and libgcc, as well as libssl at runtime, this approach results in a static libssl integrated into the routinator library (eliminating the libgcc dependency), which appears to have been the original intent. The binary itself remains dynamic as it utilized the musl provided by alpine out of the box.

This is for reference only; there's no reason to build a static routinator in this scenario as libssl is provided out of the box by alpine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,19 @@
 # -- stage 1: build static routinator with musl libc for alpine
-FROM rust:1.36.0-stretch as build
+FROM alpine:3.10.1 as build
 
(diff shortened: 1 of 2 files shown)


wk:
Closing this as the PR is for informational purposes only as an example of the requirements for a successful static libssl build.

PR #199, opened 2019-08-16
use alpine as base and build dynamic routinator in docker

wk:
Superset of #196 for a clean dynamic build with alpine as base and runtime, with openssl dependency met.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,20 @@
 # -- stage 1: build static routinator with musl libc for alpine
-FROM rust:1.36.0-stretch as build
+FROM alpine:3.10.1 as build
 
-RUN apt-get -yq update && \
-    apt-get -yq install musl-tools
-
-RUN rustup target add x86_64-unknown-linux-musl
+RUN apk add rust cargo openssl-dev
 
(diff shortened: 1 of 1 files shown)


partim:
Given that #197 drops the need for OpenSSL, how would you prefer to proceed? Keep using the official Rust build images based on Debian or switch to this new approach (with openssl-dev removed again).

I don’t feel strongly either way. I do want to make another Rust requirements jump before 1.0 to switch to async, but Alpine seems to track stable Rust releases fairly quickly, so that shouldn’t really be a problem.

wk:
Going fully alpine way seems to make sense, and should result in a slightly leaner container (everything is dynamic, but correctness is guaranteed by relying on same environment at build time and at runtime). With the switchover from openssl to rustls, the `openssl-dev` dependency at build time and/or `libgcc` dependency at runtime may also not be required.

If more granular control is ever needed later in terms of rust version, going back to using the rust container for the build will always remain an option.

partim:
Ack. I quickly remove the `openssl-dev` installation directly in your branch. Apparently I can do that and hope that’s okay.

PR #222, opened 2019-09-15
doc minor fix

bg6cq:


partim:
Thank you for the pull request! This does indeed look much better!

PR #231, opened 2019-10-19
Require /dev/log socket in systemd scripts

brada4:
Otherwise they often dont log anything after a reboot....


brada4:
some distros come with jornald only

PR #235, opened 2019-10-31
Correct path for the TAL files.

remcovz:


wk:
This seems correct. The comment must've been forgotten when the `Dockerfile` was transitioned from running the process as `root` to a configurable user.

partim:
Thank you for the pull request!

PR #239, opened 2019-11-11
Simplify CSV logic

hellerve:
This PR simplifies ther CSV logic a bit by using functionality provided by `Option` rather than matching. It’s a little verbose in one line, but IMO still quite readable.

I also found a little formatting inconsistency in repository.

If you want me to split those two things up, let me know!

Cheers

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -382,14 +382,9 @@ fn ext_csv_origin<W: io::Write>(
 ) -> Result<(), io::Error> {
     match addr.cert() {
         Some(cert) => {
-            match cert.signed_object() {
-                Some(uri) => {
-                    write!(output, "{}", uri)?;
-                }
-                None => write!(output, "N/A")?
-            }
             let val = cert.validity();
-            writeln!(output, ",{},{}/{},{},{},{}",
+            writeln!(output, "{},{},{}/{},{},{},{}",
+                cert.signed_object().map(|x| x.to_string()).unwrap_or("N/A".to_string()),
                 addr.as_id(),
                 addr.address(), addr.address_length(),
                 addr.max_length(),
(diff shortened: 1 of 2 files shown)


hellerve:
I fixed the clippy warnings introduced by me, but there seem to be a few others; I can fix them if desired?

partim:
Thank you for the pull request!

However, you are introducing extra allocations via the `to_string` calls whereas the original code simply writes what’s there directly into the output.

As for the Clippy warnings currently in master, I was a bit scared by the logic changes they imply and want to take my time to get them right.

hellerve:
AFAIK that’s not quite right, since `writeln!` will stringify anything that is not already a string anyway (I mean, it has to). If you’d hide that, we can make that happen too, of course.

As for the warnings: I totally get that. It also seems as if some of the warnings are a little spurious (since `Serial` doesn’t actually define `cmp`, only `partial_cmp`). If you want this off your mind, I can take a stab at it!

hellerve:
Just an update: I’m wrong regarding allocations, at least concerning the default argument (which gets treated at compile time to avoid intermediate allocations).

partim:
Since I also went and looked: The format string is magically converted via `format_args!` compiler built-in. The `write!` macro then calls `write_fmt` on its first argument which for `io::Write` creates a hidden impl for `fmt::Write` with calls to `fmt::Write::write_str` calling `io::Write::write_all`. There really are no hidden allocations.

partim:
As for the Clippy warnings: Thanks for the offer! I’m completely fine with silencing Clippy via `#[allow(...)]` where reasonable, but I think at least in some of the cases rewriting with a match makes sense since it more explicitly calls out the places where we fall through on purpose.

hellerve:
Cool, so I just played around with this for a bit, and there are ways to use `Option` functionality instead of `match`, but I don’t think that the long function chains (using a whole bunch of combinators) are in any way better :)

Thanks for reviewing this, I’ll close it for now! My offer to help out with anything at all still stands, though!

PR #258, opened 2019-11-29
Use a Docker volume instead of a host filesystem directory to store TALs.

ximon18:
This avoids having to fiddle with host permissions.


ximon18:
Cc @wk.

PR #261, opened 2019-12-20
Compile-time TLS library selection.

ximon18:
Based on the existing `tls-via-features` branch but with `master` merged in, `Dockerfile` updated and `Cargo.lock` regenerated (by `cargo build --release`).

"Tested' with default features via the Docker build, connecting to Krill v0.4.2-pre via NGINX with a custom CA certificate, which wasn't possible with stock Routinator.

Before:
```
$ routinator -vvv vrps -o output.json -f json --complete
https://nginx/ta/ta.cer: https://nginx/ta/ta.cer: error trying to connect: invalid certificate: UnknownIssuer
```

After:
```
RRDP: Known server https://nginx/rrdp/notification.xml at /home/routinator/.rpki-cache/repository/rrdp/2902116796
Found valid trust anchor https://nginx/ta/ta.cer. Processing.
RRDP https://nginx/rrdp/notification.xml: Updating server
Serials: us 1, them 1
RRDP https://nginx/rrdp/notification.xml: Delta update succeeded.

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,7 @@
 name: ci
-on: [push, pull_request, release]
+on: [push, pull_request]
 jobs:
   test:
-    if: github.event_name != 'release'
     name: test
     runs-on: ${{ matrix.os }}
     strategy:
@@ -23,39 +22,3 @@ jobs:
     - run: cargo build --verbose
     - run: cargo test --verbose
 
-  dockerhub:
-    name: build-and-push
-    needs: test
-    if: (github.event_name == 'push' && github.ref == 'refs/heads/master') || 
-        (github.event.name == 'release' && startsWith(github.ref, 'refs/tags/v'))
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        tls: [default-tls, default-tls-vendored, rustls-tls]
-    steps:
-    - name: Checkout repository
-      uses: actions/checkout@v1
-
-    - name: Login to Docker Hub
-      run: |
-        echo ${{ secrets.DOCKERHUB_PASSWORD }} | docker login -u ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
-
-    - name: Build and push Docker image
-      run: |
-        set -eux
-        DOCKER_HUB_REPO="nlnetlabs/routinator"
-        DEFAULT_TLS="rustls-tls"
-
-        # Determine the base tag to use, is this a push to master or the release
-        # of a tag?
-        TAG="latest"
-        [[ "${{ github.event.name }}" == "release" ]] && TAG=${GITHUB_REF##refs/tags/}
-
-        # Determine the feature extension to add to the Docker image tag.
-        # No feature extension is applied to the tag if the default feature
-        # set is being used.
-        FEATURE_EXT=""
-        [[ "${{ matrix.tls }}" != "${DEFAULT_TLS}" ]] && FEATURE_EXT="-${{ matrix.tls }}"
-        docker build -t ${DOCKER_HUB_REPO}:${TAG}${FEATURE_EXT} --build-arg 'TLS=${{ matrix.tls }}' .
-
-        docker push ${DOCKER_HUB_REPO}:${TAG}${FEATURE_EXT}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,8 +26,8 @@ jobs:
   dockerhub:
     name: build-and-push
     needs: test
-    if: (github.event_name == 'push' && github.ref == 'refs/heads/master') ||
-        (github.event.name == 'release' && startsWith(github.ref, 'refs/tags/v')
+    if: (github.event_name == 'push' && github.ref == 'refs/heads/master') || 
+        (github.event.name == 'release' && startsWith(github.ref, 'refs/tags/v'))
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,8 @@
 name: ci
-on: [push, pull_request]
+on: [push, pull_request, release]
 jobs:
   test:
+    if: github.event_name != 'release'
     name: test
     runs-on: ${{ matrix.os }}
     strategy:
@@ -22,3 +23,39 @@ jobs:
     - run: cargo build --verbose
     - run: cargo test --verbose
 
+  dockerhub:
+    name: build-and-push
+    needs: test
+    if: (github.event_name == 'push' && github.ref == 'refs/heads/master') ||
+        (github.event.name == 'release' && startsWith(github.ref, 'refs/tags/v')
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        tls: [default-tls, default-tls-vendored, rustls-tls]
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v1
+
+    - name: Login to Docker Hub
+      run: |
+        echo ${{ secrets.DOCKERHUB_PASSWORD }} | docker login -u ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
+
+    - name: Build and push Docker image
+      run: |
+        set -eux
+        DOCKER_HUB_REPO="nlnetlabs/routinator"
+        DEFAULT_TLS="rustls-tls"
+
+        # Determine the base tag to use, is this a push to master or the release
+        # of a tag?
+        TAG="latest"
+        [[ "${{ github.event.name }}" == "release" ]] && TAG=${GITHUB_REF##refs/tags/}
+
+        # Determine the feature extension to add to the Docker image tag.
+        # No feature extension is applied to the tag if the default feature
+        # set is being used.
+        FEATURE_EXT=""
+        [[ "${{ matrix.tls }}" != "${DEFAULT_TLS}"]] && FEATURE_EXT="-${{ matrix.tls }}"
+        docker build -t ${DOCKER_HUB_REPO}:${TAG}${FEATURE_EXT} .
+
+        docker push ${DOCKER_HUB_REPO}:${TAG}${FEATURE_EXT}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,6 @@ jobs:
         # set is being used.
         FEATURE_EXT=""
         [[ "${{ matrix.tls }}" != "${DEFAULT_TLS}"]] && FEATURE_EXT="-${{ matrix.tls }}"
-        docker build -t ${DOCKER_HUB_REPO}:${TAG}${FEATURE_EXT} .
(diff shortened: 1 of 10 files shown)


ximon18:
The PR has been updated to:
- Make `rustlsls` the default TLS implementation, thereby maintaining the current default.
- Support building the Docker image with different `--features XXX` values passed to `cargo build`.
- Support tagging the Docker images according to tag pattern: `nlnetlabs/routinator:<version>[-<tls type>]`.

Tagging pattern:
 where `<version>` is either `latest` for pushes to master, or `vX.Y.Z` for releases, and where `<tls type>` is the TLS feature name from `Cargo.toml`, e.g. `default-tls`, and where no `-<tls type>` will be appended to the tag in the case of feature `rustls-tls` thereby maintaining consistency with the builds pushed to Docker hub today.

Why builld & publish from GitHub Actions?
Today we build and publish to Docker Hub using Docker Hub Automated Builds.

However, those builds do not support build arguments except via a `hooks/build` command based on an environment variable which can only be set to a single value via the Docker Build Automated Builds UI.

To be able to build and publish multiple image variants we must therefore build them somewhere else. Via GitHub Actions would seem to make sense as we already use it. Additionally the proposed configuration would only build and publish any Docker images if all CI tests pass, whereas today a commit to master can cause one or more CI failures yet Docker Hub will still build and publish a new `latest` image.

NOTE: the Workflow changes have NOT been tested.

ximon18:
@partim: Can we document what the different TLS feature names mean? Are the names good enough or could they be simpler and clearer?

ximon18:
FYI I have forked routinator at https://github.com/ximoneighteen/routinator in order to test the Docker build and GitHub Actions workflow. On the first run it failed, see: https://github.com/ximon18/routinator/runs/396304763#step:4:386

wk:
@ximon18 take a look at [using environment variables for builds](https://docs.docker.com/docker-hub/builds/#environment-variables-for-builds) if you'd like to stay with Docker Hub Automated builds. In this setup, you can use a combination of tag regexps and variables to select/mark/build different versions. 

Alternatively, check the `Repository Links` feature which will allow you to easily configure a 2nd Docker Hub repository (with a different, Dockerfile) to build a routinator-feature:version in parallel whenever a change to the primary repository is detected.

Finally, you can indeed use build hooks which could consume the auto-generated variable resulting from tagging to customize build outcome.

Moving to GitHub Workflows is also an option. In either case, having a variant ARG is a clean way to accomplish the desired result, the only variance is in how it gets populated.

ximon18:
Thanks @wk. Perhaps I'm missing something but ...

> @ximon18 take a look at using environment variables for builds if you'd like to stay with Docker Hub Automated builds. In this setup, you can use a combination of tag regexps and variables to select/mark/build different versions.

I do not see how we can stay with Docker Hub repo and continue to use a single repo as it only supports a single environment variable for passing to the build and we need to build multiple different images which would need multiple different values of a single environment variable. I'm also not sure I like the idea of using an env var as env vars persist into the final run time container, while this var is only intended to affect the Dockerfile during build time.

> Finally, you can indeed use build hooks which could consume the auto-generated variable resulting from tagging to customize build outcome.

I cannot see how to do this using the single env var that can be set and a single build flow, I'd have to encode multiple values into the env var value and split them in a `build` hook, or code them into the build hook itself, and even then I'd have to build multiple images from a build process intended to build a single image which seems unwise, I wonder even if Docker Hub would correctly show all of the built tags in that case.

> Alternatively, check the Repository Links feature which will allow you to easily configure a 2nd Docker Hub repository (with a different, Dockerfile) to build a routinator-feature:version in parallel whenever a change to the primary repository is detected.

I was not familiar with `Repository Links`, thanks! That's an interesting idea but then the Routinator images would be spread over multiple Docker Hub repos and end users would have to use the right repo name rather than a tag variant, which is not how other images do it, e.g. https://hub.docker.com/_/python has many different tags all in a single repo, including the `slim` tags which could be considered comparable to the idea of multiple Routinator builds using different TLS implementations.

> Moving to GitHub Workflows is also an option. In either case, having a variant ARG is a clean way to accomplish the desired result, the only variance is in how it gets populated.

Which leaves me with this option which this PR attempts to implement, though currently it fails at the first hurdle as it seems a dependency must be added to the Dockerfile to make the `default-tls-vendored` build work. See: thread 'main' panicked at 'don't know how to configure OpenSSL for x86_64-alpine-linux-musl', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-src-111.6.0+1.1.1d/src/lib.rs:178:18

Once again, thanks for the awesome input, v appreciated.

wk:
@ximon18 

You're correct in that unfortunately Docker Hub Automated Builds don't allow defining an environment variable on a per-rule basis, only on a per-build basis. 

The easiest way to accomplish what is desired above is probably with two `Dockerfile` instances in the repository, each configured with the correct argument per-populated by default, then something like this should work and build 2 instances at each trigger:

image

For build hooks, since the build hook allows uploading to a *different* repository at the end of a successful build, the workaround there is to create a second Docker Hub repository, configure a parallel build (with the correct variables passed), and then use the hooks to upload the built image with the correct name to the first repository (resulting in one repository with multiple variant tags). This can be used with or without `Repository Links`. Not by any means clean, but should be functional.

ximon18:
Thanks @wk. In this case it's not two images but 3 (one each for TLS implementations default-tls, default-tls-vendored, rustls-tls - though I'm going to talk to @partim to understand what exactly these are), so then that's three Dockerfiles (which cannot inherit from each other, unless we publish a base image they build on which is then another Dockerfile) and three Docker Hub repositories, which feels like it's getting out of control to me. Is my understanding correct?

partim:
I _think_ we don’t need the `default-tls-vendored` – it means that the `openssl` crate will use its own included vendored version of OpenSSL. Since the system OpenSSL works for our builds, we might as well use that and make the build quicker.

ximon18:
>  think we don’t need the default-tls-vendored – it means that the openssl crate will use its own included vendored version of OpenSSL. Since the system OpenSSL works for our builds, we might as well use that and make the build quicker.

Okay, then that leaves two variants, `rustls-tls` which is the current default and then system OpenSSL. I propose:
- Remove the `default-tls-vendored` feature.
- Adopt the two Dockerfile approach suggested by @wk.
It's unfortunate that we then end up with a lot of overlap between the two Dockerfiles but perhaps it is better than one Dockerfile that tries to install different dependencies and use different build rules according to a control argument.

@partim: Are you okay with this?

partim:
Do we need to keep those Dockerfiles in the root directory or can we move them somewhere – or at least one of them?

wk:
The files can live anywhere within the repository.

partim:
Them maybe we can keep the regular Dockerfile in the root and have the secondary one in, dunno, `etc`?

ximon18:
> Them maybe we can keep the regular Dockerfile in the root and have the secondary one in, dunno, etc?
`etc` is typically associated with operating system configuration files, so that feels a little strange to me, especially as then why is one Dockerfile in `etc` and the other not?

I've often seen more than one Dockerfile in the root, as long as it's just one more (`Dockerfile.tls`) it won't impact you much so I'd just put it in the root. If we end up needing more later we can revisit the structure.

Alternatively you could create a `docker` subdirectory and put all the Docker stuff in there, that's what we did for the Internet.nl project. For Krill there is also a `docker` subdirectory but only for the additional files copied into the image, the `Dockerfile` is still in the root.

partim:
Let’s do that and put all of them into a `docker` directory.

ximon18:
> Let’s do that and put all of them into a docker directory.
I'll update the PR.

ximon18:
The PR has been updated to:
- Move the default Dockerfile into a new `docker` subdirectory.
- Add a second similar Dockerfile called `docker/Dockerfile.tls` which builds Routinator with OpenSSL (default-tls) support instead of Rustls support.
- Added a new section to the README on how to build/run with default-tls instead of Rustls.
- Removed the GitHub Actions workflow modifications as they are no longer needed.

Still to do:
- If we accept this the Docker Hub automated build rules have to be amended to:
-- Build and tag as usual from `docker/Dockerfile` in the `/` context.
-- Build and tag as `default-tls` from `docker/Dockerfile.tls` in the `/` context.
- Amend the crate publishing process to ensure that `default-tls` can be used via `cargo install` as currently attempting to do so results in an error do to the existing crates lacking the necessary features.

The two new Dockerfiles have been tested with a local copy of the Krill E2E test which confirms that Routinator works in both variants and as expected the Rustls variant fails with a certificate error while the default-tls variants passes the tests.

Question: does the `socks` feature really need to be specified by hand as well as `default-tls` as is done in this PR (based on the original tls-with-features branch)?

partim:
I think it does. It is being invoked as part of the default features but those are being disabled when you switch away from `rustls-tls`.

ximon18:
After discussing with @partim this PR has been updated to rename `default-tls` to `native-tls`. At this time a `native-tls` specific crate will not be published and the `cargo build` instructions in README.md are fine for now.

Remaining actions:
- [ ] Final review of this PR.
- [ ] Adjust the Docker Hub automated build rules *before* merging this PR so that the default Dockerfile build does not fail (due to the Dockerfile being moved into the docker subdirectory):
  - [ ] Edit the default build rule to point to `docker/Dockerfile`.
  - [ ] Add a new build rule to build `docker/Dockerfile.native-tls` and tag it as `:native-tls` and `:vX.Y.Z-native-tls`.
- [ ] Merge this PR.
- [ ] Update README content at readthedocs and crates.io if not automatically updated.

ximon18:
Docker Hub automated build rules configuration *before* applying changes:
image

ximon18:
I verified again with a local copy of the Krill E2E test building with the main Dockerfile in the new location and the new `native-tls` Dockerfile.

ximon18:
Merging as agreed with @partim.

ximon18:
The Krill E2E test framework has been updated to use the new nlnetlabs/routinator:native-tls image and passed its self check. See: https://github.com/NLnetLabs/rpki-deploy/runs/413461034

ximon18:
@partim / @wk: Although this PR is closed, for completeness I just want to add something here.

The dual Dockerfile approach defined in this PR is working. If, however, the second Dockerfile is ever considered to be too annoying to maintain (as it mostly a duplicate of the first), I think I just realized a some ways that *might* make this possible with a single Dockerfile.

1. It seems that [Docker Hub Autobuild sets a `DOCKER_TAG` environment variable](https://docs.docker.com/docker-hub/builds/advanced/) during the build. As the new Autobuild rule uses a separate tag containing `native-tls` it *might* be possible to use the [*"Conditions using build arguments"* approach](https://medium.com/@tonistiigi/advanced-multi-stage-build-patterns-6f741b852fae) but based on env vars instead of build args to define two base stages and only use the output of one of them. As only the first stage of the two Dockerfiles differ this would allow for a common third stage. There would however still be a lot in common between the first two stages.

2. Alternatively, somewhat hacky, the arguments to `apk add` and to `cargo build` that vary between the two builds could be made conditional using shell conditional logic inside the Docker `RUN` value. 

However, I'm inclined to think that the dual Dockerfile approach is more readable than using multi-build stage or shell tricks within a Dockerfile to implement conditional logic that Docker itself does not offer.

PR #262, opened 2019-12-20
Disable apk caching to reduce image size

ximon18:
Avoid storing ~1.4M of /var/cache/apk/ files in the image which are not needed. Pkg installation still works without them.


ximon18:
Cc: @wk

wk:
Makes perfect sense to me (can also be done at the clean-up stage as `apk cache clean` as well).

PR #263, opened 2019-12-24
Fix description in prometheus endpoint

ties:
Nitpick: In the prometheus endpoint the description of the RRDP duration words it as rsync duration.


partim:
Oops! Thanks for the PR and the fix!

PR #273, opened 2020-02-17
dd bird and bird2 compatible outform

### LINKED ISSUE #272: Extend outform with bird1 and bird2 format
It would be cool if routinator's -f option allows bird1 and bird2 format so that users of the software can immediately plug the output into the bird1 and bird2 daemon. (#31)

The bird1 format is as following:
```
# IPv4
roa 100.128.0.0/9 max 24 as 21928
roa 78.192.0.0/10 max 11 as 12322
```
```
# IPv6
roa 2404:8d00:1c00::/38 max 40 as 24378
```

The bird2 format is as following:
```
# IPv4
route 100.128.0.0/9 max 24 as 21928
route 78.192.0.0/10 max 11 as 12322
```
```
# IPv6
route 2404:8d00:1c00::/38 max 40 as 24378
```

netravnen:
This fixes #272 (#32)

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -39,6 +39,16 @@ pub enum OutputFormat {
     /// Specifically, this produces as `roa-set`.
     Openbgpd,
 
+    /// BIRD configuration format.
+    ///
+    /// Specifically, this produces as `roa table`.
+    Bird1,
+
+    /// BIRD2 configuration format.
+    ///
+    /// Specifically, this produces as `route table`.
+    Bird2,
+
     /// RPSL output.
     ///
     /// This produces a sequence of RPSL objects with various fields.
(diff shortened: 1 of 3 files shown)


partim:
Thank you for the PR!

I am on vacation for the rest of the month and will review and merge as soon as I am back.

src/output.rs
```diff
@@ -483,6 +501,62 @@ fn openbgpd_footer<W: io::Write>(
 }
 
 
+//------------ bird1 ------------------------------------------------------
+
+fn bird1_header<W: io::Write>(
+    _vrps: &AddressOrigins,
+    output: &mut W,
```
partim: As suggested by Clippy, the unused `output` should be changed to either `_output` or even just `_` on all these occurrences.

partim:
I have replaced this PR with #290, which only fixes the warnings and makes it compatible with then new HTTP server.

PR #277, opened 2020-03-11
Enable support for CTRL-C from outside the container.

### LINKED ISSUE #257: Support CTRL-C of Dockerised Routinator when run in foreground
As it doesn't respond to received signals, e.g.:

```
root@routinator:~# docker run --rm -v /etc/routinator/tals:/home/routinator/.rpki-cache/tals nlnetlabs/routinator
^C^C^C^C^C^C
^Zbg
^Z^Z^Z^Z^Z
```

See: https://docs.docker.com/v17.12/engine/reference/run/#foreground

ximon18:
Fixes #257. See https://github.com/krallin/tini for Tini documentation.
The "proper" fix is to handle the signal sent to the process, though Tini also handles other "pid 1" responsibilities as well (e.g. reaping finished background processes).


ximon18:
@wk: any thoughts on this?

wk:
There's a pretty clean manner to handle this in-process in Rust [documented here](https://piers.rocks/rust/signal/sigint/container/docker/2019/01/06/rust-catch-ctrl-c-container.html).

If you do wish to go the tini route, a better way in this particular scenario might be to rely on the [alpine package](https://github.com/krallin/tini#alpine-linux-package) (for backwards compatibility with docker priort to 1.13, if this is relevant), or to rely on the [built-in tini](https://github.com/krallin/tini#using-tini) by updating the documentation for container launch.

ximon18:
Hi @wk,

Thanks for the advice. I'm not a Rust engineer myself so I'll let @partim process the Rust in-process handling suggestion. I also wasn't aware of the Alpine package for `tini`, thanks!

I'm not a fan of the `--init` Docker native support for using `tini` as it requires the end user to do yet another thing to run Routinator via Docker to get behaviour which is (IMO) expected to just work out-of-the-box.

@partim, do you have a preference here? Documentation over baked-in default behaviour? As it stands currently you can CTRL-C Routinator when using the Rust binary directly, but not when running via Docker unless you remember to pass `--init` to Docker when running Routinator.

Ximon

partim:
Ah, darn, I thought it was clear that #257 was meant for me to add that to Routinator itself. Since it already handles SIGUSR1 for restarting validation, this isn’t even all that difficult to add.

partim:
So, erm, this is all a bit more tricky than I though. The issue here is the forked rsync processes. Apparently, to do the container things properly I also need to clean those out properly – sending them a SIGTERM and then waitpid-ing for them. Doing all that upon a signal significantly complicates the code, so I think leaving this to `tini` might be a good idea after all.

ximon18:
>  Doing all that upon a signal significantly complicates the code, so I think leaving this to tini might be a good idea after all.

Adding Docker specific behaviour to the application doesn't seem right when there's a Docker specific solution out there already. So yes let's go with Alpine tini. I've synced the PR with latest master.

PR #278, opened 2020-03-11
Remove native TLS Docker image.

ximon18:
As discussed with @partim, native TLS is available to those who want to build support for it but the only user of the native TLS image known to date can address their use case via the existing Routinator support for specifying a custom CA cert. If we merge this to master we should **first** disable the build rule at Docker Hub. We could also optionally move `Dockerfile` back to the `/` and remove the then empty `docker` subdirectory, which would also require editing the build rule at Docker Hub to find the `Dockerfile` in the right place.


ximon18:
Force pushed to replace the accidentally overly long commit msg with a shorter one.

partim:
@ximon18, would you mind adding a commit moving the sole remaining Dockerfile back to the root as suggested?

I’ll be fine with merging after that. Perhaps you can do that in sync with the necessary Docker Hub updates?

ximon18:
@partim: I'll take care of it.

ximon18:
Done. I have also removed the `native-tls` specific build rules at Docker Hub.

I have *NOT* yet edited the Docker Hub build rule for the remaining Dockerfile telling Docker Hub to find it now in the project root as that should be done immediately prior to merge of this branch to prevent breaking builds caused by other commits to master prior to merge.

ximon18:
@partim: Ping me when you're ready to merge this and I'll adjust the Docker Hub build rule.

partim:
Feel free to merge at any time.

ximon18:
The `nlnetlabs/routinator:native-tls` Docker Hub tag has been deleted.

PR #297, opened 2020-04-20
Catch SIGINT and exit.

### LINKED ISSUE #257: Support CTRL-C of Dockerised Routinator when run in foreground

partim:
This is primarily for use inside containers.

Caveat: It will only work in server mode.

Fixes #257

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -30,6 +30,9 @@ New
   Validator. ([#291])
 * The _csvext_ output format is now also available via the HTTP server at
   the `/csvext` path. ([#294])
+* On Unix, Routinator in server mode now also catches SIGINT and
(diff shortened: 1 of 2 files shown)


partim:
Turns out, just exiting on SIGINT isn’t enough. Proper handling requires lots of rewiring. We’ll got with #277 instead.

PR #311, opened 2020-04-23
Prepare for 0.7.0-rc1

### LINKED ISSUE #302: Add --force as init option
The init section of the man page now reads:

> -f     Forces installation of the TALs even if the TAL directory already exists.

add `--force`, because that works too.

### LINKED ISSUE #306: detach flag in server mode - man page
The `server` sub-command can detach using both `-d` and `--detach`. Only the latter is documented in the man page.

### LINKED ISSUE #307: Typo in tal-label description
If the options missing or if a TAL isn't mentioned in the **optio**...

### LINKED ISSUE #310: /csvext missing from HTTP Service
AFAIK the HTTP server can now do csvext, but the option isn't documented in the HTTP Service section of the man page.

partim:
Fixes #302. Fixes #306. Fixes #307. Fixes #310.

### DIFF
--- a/doc/routinator.1
+++ b/doc/routinator.1
@@ -1,4 +1,4 @@
-.TH "routinator" "1" "April 34, 2020" "NLnet Labs" "routinator 0.7.0-rc1
+.TH "routinator" "1" "April 23, 2020" "NLnet Labs" "routinator 0.7.0-rc1
 .\"
 .\" routinator.1 -- RPKI Relying Party software
 .\"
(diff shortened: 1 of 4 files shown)


doc/routinator.1
```diff
@@ -1,4 +1,4 @@
-.TH "routinator" "1" "November 29, 2019" "NLnet Labs" "routinator 0.6.4
+.TH "routinator" "1" "April 34, 2020" "NLnet Labs" "routinator 0.7.0-rc1
```
AlexanderBand, line 1: April 34!

gthess: That's the best day of April

PR #317, opened 2020-04-27
Remove UTF8 characters from example config

devicenull:
The example configuration file contains a few UTF-8 `’` characters, rather then the plain ASCII `'` character.  Since these are only in comments, remove them to prevent issues where UTF8 gets mangled and routinator refuses to start up


partim:
I’ll spare you the ranting about it being 2020 and admit that there shouldn’t be non-ASCII characters in these files. Thanks for the PR!

PR #331, opened 2020-05-14
Update AFRINIC TAL from rsync to HTTPS

AlexanderBand:


partim:
Can you include both URLs? Routinator will prefer the HTTP one, so the order doesn’t quite matter.

PR #337, opened 2020-05-17
bump alpine to 3.11.6 in Dockerfile

wk:
Bump alpine to 3.11.6 in Dockerfile. Manually validated.


partim:
Thank you for keeping an eye on these things! That is much appreciated!

PR #339, opened 2020-06-08
FIX: dubios -> dubious

ximon18:
This PR fixes the following issue:
```
$ routinator --version
Routinator 0.7.0

$ routinator --allow-dubious-hosts --tal-dir=/tmp/tal --rrdp-root-cert=/home/ximon/src/krill/generated_cert/issuer.crt -vvv vrps
error: Found argument '--allow-dubious-hosts' which wasn't expected, or isn't valid in this context
	Did you mean --allow-dubios-hosts?

USAGE:
    routinator --allow-dubios-hosts

For more information try --help
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -259,7 +259,7 @@ impl Config {
              .takes_value(true)
         )
         .arg(Arg::with_name("allow-dubious-hosts")
-             .long("allow-dubios-hosts")
+             .long("allow-dubious-hosts")
              .help("Allow dubious host names in rsycn and HTTPS URIs")
         )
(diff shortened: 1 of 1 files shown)


partim:
Oops.

I’d want to keep the wrong spelling in as a quiet alternative (i.e., it’s there but not documented and doesn’t show up in the usage output either) to avoid having a breaking change before the next intentional breaking release.

I’ll have a look into how to achieve this and then push another commit to this PR.

partim:
That was easy. Clap is great!

PR #349, opened 2020-06-11
bump alpine to 3.12.0 in Dockerfile

wk:
Manually validated.


partim:
Thank you for the PR! It will get merged after 0.7.1 is released.

PR #352, opened 2020-06-12
Debian packaging tweaks

ximon18:
I'm not that familiar with Debian packaging so these changes might be rubbish, but on reading [cargo-deb docs](https://github.com/mmstick/cargo-deb) and [Debian packaging guidelines](https://www.debian.org/doc/manuals/debian-faq/pkg-basics.en.html) I made the following tweaks:

- Adds conf-files per cargo-deb docs to avoid routinator.conf being overwritten on upgrade.
- Changes section from "rust" to "net" as the implementation language is less descriptive of the tool than that it is a networking tool.
- Removes license-file as cargo-deb docs say it will already use the existing "license" field from Cargo.toml.

@wk: as the original contributor of the cargo-deb metadata in commit e828695a7f83c745f23845df79e1e41c1bb19e85, and @rfc1036 as the maintainer of https://salsa.debian.org/md/routinator/-/tree/master/debian, I'm curious to know what you think of these changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ AS number will be the origin of BGP route announcements for it. \
 Routinator is a RPKI relying party software written in Rust. """
 depends = "$auto, rsync"
 section = "net"
-priority = "extra"
+priority = "optional"
 assets = [
     ["target/release/routinator", "usr/bin/", "755"],
     ["README.md", "usr/share/doc/routinator/", "644"],
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ tls = []
 
 [package.metadata.deb]
 maintainer = "The NLnet Labs RPKI Team <rpki@nlnetlabs.nl>"
+license-file = ["LICENSE", "0"]	
 extended-description = """\
 The Resource Public Key Infrastructure provides cryptographically signed \
 statements about the association of Internet routing resources. \
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,6 @@ tls = []
 
 [package.metadata.deb]
 maintainer = "The NLnet Labs RPKI Team <rpki@nlnetlabs.nl>"
-license-file = ["LICENSE", "0"]
 extended-description = """\
 The Resource Public Key Infrastructure provides cryptographically signed \
 statements about the association of Internet routing resources. \
@@ -64,8 +63,8 @@ In particular, it allows the holder of an IP address prefix to publish which \
 AS number will be the origin of BGP route announcements for it. \
 Routinator is a RPKI relying party software written in Rust. """
 depends = "$auto, rsync"
-section = "rust"
-priority = "optional"
+section = "net"
+priority = "extra"
 assets = [
     ["target/release/routinator", "usr/bin/", "755"],
     ["README.md", "usr/share/doc/routinator/", "644"],
@@ -75,6 +74,9 @@ assets = [
     ["etc/routinator.conf.system-service", "etc/routinator/routinator.conf", "644"]
 ]
 maintainer-scripts = "debian"
+conf-files = [
+    "etc/routinator/routinator.conf"
+]
 
 [package.metadata.deb.variants.minimal]
 assets = [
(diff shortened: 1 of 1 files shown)


rfc1036:
No real routinator Debian or Ubuntu package will be possible until all dependencies will be available in Debian anyway.
Without all the integration that I have implemented in my package (configuration files with sane paths, sandboxing, depending on rpki-trust-anchors, etc...) then I do not think that this brings much value over just copying the routinator binary.

(Extra priority is deprecated and should not be used anymore.)

ximon18:
Thanks for the very fast reaction @rfc1036  !

> No real routinator Debian or Ubuntu package will be possible until all dependencies will be available in Debian anyway.

Forgive my ignorance, but if I run `cargo deb` on my Ubuntu 18.04 system I can then do `apt install ./target/debian/routinator_0.7.1~rc2-bis_amd64.deb` and end up with `/usr/bin/routinator` which seems to work fine.

Is the issue for a "real" package to do with not depending on stock crates.io downloads but on verified build dependencies or something?

> Without all the integration that I have implemented in my package (configuration files with sane paths, sandboxing, depending on rpki-trust-anchors, etc...) then I do not think that this brings much value over just copying the routinator binary.

Understood. I came across this while looking into packaging for Krill rather than Routinator so really just wanted to understand if the choices made in this configuration are sane or not, and am aware that you have put a lot of effort into proper Debian packaging of Routinator.

> (Extra priority is deprecated and should not be used anymore.)

That's useful to know, thanks. Is that documented somewhere?

Thankyou very much!

rfc1036:
For very good reasons Debian (and Ubuntu) packages must be built only from dependencies available in the Debian archive, and no network access is allowed.

ximon18:
That only applies to official packages right?

wk:
On value:
Providing metadata for `cargo-deb` carries very little cost and currently allows locally building a working package that's been successfully tested on recent versions of Debian and Ubuntu.

It enables both clean local installation (where any subsequent removal and updating will be handled by the package manager), as well as local distribution within an organization from an internal repository. The resulting generated package being unsuitable for inclusion in upstream Debian or Ubuntu repositories does not hinder either of these use cases.

On the change set:
I can successfully build a package from the `deb-pkg-tweaks` branch, and the changes seem reasonable - however, as @rfc1036 pointed out, I believe "optional" is, indeed, the correct priority.

wk:
> That's useful to know, thanks. Is that documented somewhere?

This is outlined [here](https://www.debian.org/doc/debian-policy/ch-archive.html#s-priorities).

ximon18:
> > That's useful to know, thanks. Is that documented somewhere?
> 
> This is outlined [here](https://www.debian.org/doc/debian-policy/ch-archive.html#s-priorities).

Thanks, PR updated accordingly.

ximon18:
Regarding:
> Removes license-file as cargo-deb docs say it will already use the existing "license" field from Cargo.toml.

Actually we should revert the license-file change that I proposed because the [Debian Policy manual](https://www.debian.org/doc/debian-policy/ch-docs.html#copyright-information) says:
> Every package must be accompanied by a verbatim copy of its copyright information and distribution license in the file /usr/share/doc/package/copyright. This file must neither be compressed nor be a symbolic link.
> 
> In addition, the copyright file must say where the upstream sources (if any) were obtained, and should include a name or contact address for the upstream authors. This can be the name of an individual or an organization, an email address, a web forum or bugtracker, or any other means to unambiguously identify who to contact to participate in the development of the upstream source code.

With `license-file` the verbatim license is included, but not so when `license-file` is omitted, instead you only get a license reference like so (from the Krill project):
```
Upstream Name: krill-ubuntu1804
Source: https://github.com/NLnetLabs/krill
Copyright: The NLnet Labs RPKI team <rpki-team@nlnetlabs.nl>
License: MPL-2.0
```
So, I think I'll revert the license-file change part of this proposal.

ximon18:
After discussion with @partim I believe this PR can be merged as the only changes left after review and edits are uncontroversial, change section from "rust" to "net" and mark the configuration file so that it won't be automatically overwritten on upgrade.

PR #360, opened 2020-06-29
Deb packaging automation

ximon18:
Based on the work done for NLnetLabs/Krill#261:
- Builds for Ubuntu 16.04, 18.04, 20.04 and Debian 9 and 10. By configuration only two variants are built: modern systemd and minimal systemd, but differences in available standard library (e.g. glibc) versions available per platform when building with $auto depends requires that each package be built on the target platform to get the correct `Depends:` entry in the created DEB package. This is considered less brittle than hand specifying a lowest common dependency set in Cargo.toml for cargo-deb.
- Includes Routinator man page.
- Creates a routinator user and a systemd (disabled) service with a Routinator config file.
- Assumes that the target system has ca-certificates installed already.
- Use GH caching to avoid repeat compilation of unchanging cargo deb and Routinator dependencies.
- Build in Docker containers as GH runners have too many non-standard packages and libraries pre-installed.
- Sanity check the created DEBs in targeted O/S versions via LXC/LXD containers (for systemd support).

@partim : currently packaging occurs for every commit, we may want to limit that.


ximon18:
I think I should incorporate the improvements being made in https://github.com/NLnetLabs/krill/pull/276 to this PR.

ximon18:
So, the package seems to work but there are two important things to know:

1. Post installation the user must run a command like this to 'init' routinator correctly as the correct user using the correct config file:
```
sudo -u routinator routinator --config /etc/routinator/routinator.conf init --accept-arin-rpa
```

2. For the package to run routinator as the correct user on older platforms (Ubuntu 16.04, Ubuntu 18.04, Debian 9) the minimal service file has been updated to include `User=routinator`, meaning it is not as minimal as it was. We could add a third service file to avoid altering the minimal one that we have today. _Note: I do NOT use the Routinator `user` and `group` configuration options to drop privileges from user root as they have no effect if `--detach` is not also used, but when using `--detach` the systemd unit no longer works as without further configuration (`Type=forking`, specifying the pid file, etc) systemd doesn't expect the program to fork itself and just sees the parent process exit (at least I think that's what I saw in my testing)._

Am example of the above working can be seen in the log output of the [GitHub Actions run here](https://github.com/NLnetLabs/routinator/runs/992727345?check_suite_focus=true#step:11:95).

@partim, @AlexanderBand: Any thoughts on the above?

partim:
For 1., I suggest we fix this in a separate PR.
For 2., I made #368; we can adjust the included configs as part of that.

I would like to hold off merging this PR until after #371, which is kind of big and I rather not have to solve loads of conflicts on that one.

netravnen:
## Idea plucked from a tree

Does it make sense to [test code](https://go.netravnen.de/diz26im) on e.g. the [upcoming Debian](https://www.debian.org/releases/testing/releasenotes) branch?

partim:
That seems reasonable. Let me add that, indeed.

partim (approved):

.github/workflows/pkg.yml
```diff
@@ -0,0 +1,331 @@
...
+      CARGO_DEB_VER: 94ba8f3
+      # A Routinator version of the form 'x.y.z-bis' denotes a dev build that is
+      # newer than the released x.y.z version but is not yet a new release.
+      NEXT_VER_LABEL: bis
+    name: deb-pkg
+    runs-on: ubuntu-latest
+    # Build on the oldest platform we are targeting in order to avoid
+    # https://github.com/rust-lang/rust/issues/57497. Specifying container
+    # causes all of the steps in this job to run inside a Docker container.
+    container: ${{ matrix.image }}
```
ximon18, line 50: I can't think of a way to make this expression handle the conversion from debian:11 to debian:bullseye. Without that we either have all numeric and one inconsistent named "bullseye" version due to the not-yet-present debian:11 image tag on Docker Hub, or perhaps we should switch all of the image values to name based instead of numeric based references to make them consistent. However, I at least, find numeric values easier to reason about than having to remember which image name is which version.

netravnen: > perhaps we should switch all of the image values to name based instead of numeric based references to make them consistent

https://hub.docker.com/_/debian

Personally, I rather use the normal release names and include a comment with the release number (or [just a link](https://www.debian.org/releases/) to external documentation). Rather than the numbering scheme. Makes targeting OS releases other than the stable releases way easier.

ximon18: Perhaps it's a personal preference thing, e.g. I run Ubuntu 18.04, I can't tell you what the release name is, I have no idea without looking it up. However, I think in this case I would rather have consistently named jobs and so may be forced to use release names. I'll give it some thought and return to this tomorrow.

ximon18:
Issues that still need to be resolved:
- Inconsistent job naming and workflow logic for Debian 11 due to lack of Docker Hub image numeric version tag, only `debian:bullseye` exists, not `debian:11`.
- Debian 9 package test is failing.
- Older O/S (Ubuntu 16.04 and Debian 9) package builds log error `date: unrecognized option '--rfc-email'` without causing build failure.

PR #362, opened 2020-07-02
Run ci workflow on branch or PR, not both.

ximon18:
E.g
image

See: https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,11 @@
 name: ci
-on: [push, pull_request]
+on:
+  push:
+    branches: 
+      - master
+  pull_request:
+    branches: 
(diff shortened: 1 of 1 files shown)


ximon18:
image
Looks good!

partim:
Will this also run the action when a new commit is pushed to the branch? I suspect it does via the `synchronize` pull request event action?

ximon18:
> Will this also run the action when a new commit is pushed to the branch? I suspect it does via the `synchronize` pull request event action?

The action will run either when there is a push to the PR branch or to the master branch (either directly or when the PR is merged to master).

partim (approved):

PR #377, opened 2020-09-10
Filter all VRPs intersecting with resoures from invalid CAs.

### LINKED ISSUE #370: Differences between vrps_total values in metrics file and AS/prefix pairs in csv file
Hi,
can u please explain the following differences:

While monitoring the Routinator metrics I found differences between the vrps values in the metrics file and the line count [grep on tals] in the csv file.

As far a I understand a vrp (validated roas payload) value is a one-to-one mapping between an ASN and an ip address and the csv file is an export of these mappings. 
So counting the lines in the csv file should bring up the same values, or not !?

```
VRPS per TAL info from routinator.status file :
vrps-per-tal: afrinic=1375 ripe=95178 apnic=49097 lacnic=11045 arin=17981 
-----------------

VRPS per TAL info from routinator.metrics file :
# HELP routinator_vrps_total total number of VRPs seen
# TYPE routinator_vrps_total gauge
routinator_vrps_total{tal="afrinic"} 1375
routinator_vrps_total{tal="ripe"} 95178
routinator_vrps_total{tal="apnic"} 49097
routinator_vrps_total{tal="lacnic"} 11045
routinator_vrps_total{tal="arin"} 17981
-----------------

Line grep and count on routinator.csv :
[afrinic] :  1359
[ripe]    :  95178
[apnic]   :  49059
[lacnic]  :  10057
[arin]    :  15861
```

Frank

### LINKED ISSUE #380: SLURM filtering diagnostics
I couldn't find any metrics endpoint or log message regarding the status of the exceptions currently active.  
In our scenario we have a hand full of routinator instances running globally and it would be great if we could somehow monitor  if and when they implemented the exceptions we specified.

Cloudflares gortr for instance logs the following message:
`gortr_1     | time="2020-09-21T07:00:53Z" level=info msg="Slurm filtering: 180178 kept, 1 removed, 0 asserted"`
on every reevaluation.

I would suggest to add 2 gauge metrics representing a count of `prefixFilters` and `prefixAssertions`, e.g.:
```
routinator_slurm_prefix_filters 0
routinator_slurm_local_assertions 0
```

partim:
This PR introduces super-strict filtering of invalid ROAs. It is very much a draft for now awaiting feedback.

It is based on the observation that filtering individual ROAs may lead to legitimate route announcements being dropped because ROAs only contain a single originating AS. As a consequence, a prefix possibly being announced by multiple ASs needs to be authorized via multiple ROAs. If the ROA for the AS currently announcing the prefix gets dropped for whatever reason, the prefix becomes invalid and its route gets dropped.

As an immediate measure, it was proposed to drop all objects published by a CA that has any invalid objects whatsoever. This will make all ROAs published by this CA disappear but also all child CAs. While this will solve the above issue for most practical cases, there is still a possibility that a prefix is delegated to multiple CAs which independently publish ROAs authorising different ASs. If only the ROAs published by one of those CAs is dropped, the result may again be a falsely dropped route. The only way to avoid this is to drop all authorisations (i.e., VRPs) for all address resources delegated to the invalid CA.

However, this is still not quite enough. If the VRP for a legitimate route is dropped because of above filtering and there is a less specific VRP with a max-length not covering the route, then the route suddenly becomes invalid too. To avoid that, we need to filter all VRPs that overlap with any of the resources of invalid CAs.

This is what this PR does: It creates a list of all the resources from all the CAs that had to be dropped because of issues. When creating the final set of VRPs, it filters out all VRPs whose address prefixes overlap with any of these resources, making these resources guaranteed to be RPKI unknown.

There is a new metric `routinator_vrps_unsafe` in Prometheus output or `unsafe-vrps` in status output that counts the VRPs filtered that way. At the time of writing, around 3300 VRPs or 1.9 % were filtered this way.

The question here is, whether this aggressive filtering will improve the overall security of the system. This kind of filtering can – at least in theory – be used by an attacker to actively push their target space to RPKI unknown as an initial step of a route hijack. I _think_ this is relatively difficult to achieve and the risk subsequently is very low, but I’d love to hear opinions – particularly arguments against this filtering strategy.

*  *  *

The PR also introduces new metrics: Fixes #370. Fixes #380.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -428,9 +428,9 @@ dependencies = [
 
 [[package]]
 name = "getrandom"
-version = "0.1.14"
+version = "0.1.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7abc8dd8451921606d809ba32e95b6111925cd2906060d2dcc29c070220503eb"
+checksum = "fc587bc0ec293155d5bfa6b9891ec18a1e330c234f896ea47fbada4cadbe47e6"
 dependencies = [
  "cfg-if",
  "libc",
@@ -697,11 +697,12 @@ dependencies = [
 
 [[package]]
 name = "miniz_oxide"
-version = "0.4.1"
+version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4d7559a8a40d0f97e1edea3220f698f78b1c5ab67532e49f68fde3910323b722"
+checksum = "c60c0dfe32c10b43a144bad8fc83538c52f58302c92300ea7ec7bf7b38d5a7b9"
 dependencies = [
  "adler",
+ "autocfg",
 ]
 
 [[package]]
@@ -1098,7 +1099,7 @@ dependencies = [
 [[package]]
 name = "rpki"
 version = "0.9.2-bis"
-source = "git+https://github.com/NLnetLabs/rpki-rs.git?branch=more-block-methods#fd2f7aeea688dd3a63ff713f3c70946b887b6add"
+source = "git+https://github.com/NLnetLabs/rpki-rs.git?branch=more-block-methods#be21a98f2f7770641310b367c05fa672561be088"
 dependencies = [
  "base64",
  "bcder",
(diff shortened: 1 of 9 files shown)


timbru:
It would be wise to make an exception for failures at a TA level - e.g. the MFT of a TA is not re-signed in time. Otherwise a failure at the apex of any of the TAs could wipe out all VRPs.

Note that many RIRs use an offline TA, and an online CA with the same resources just under it. So this exception may also need to be applied to this secondary layer. Alternatively it could be done based on a cap of the amount of resources (like '0/0').

Theoretically it could also use different strategies for each TA - knowing the TA set up - however, this will be cumbersome and possibly error prone / not future proof.

partim:
I guess just not filtering 0/0 should be fine, given that effectively this can only happen at a TA and all CAs that the TA fully delegates to.

timbru:
RIPE NCC has one more layer you may want to consider. They sub-delegate from "all resources" to "all ripe ncc resources":
https://www.ripe.net/manage-ips-and-asns/resource-management/certification/ripe-ncc-rpki-trust-anchor-structure

In practice however, I think it's unlikely that the "all resources" CA is ok, and the "all ripe ncc resources CA" is not.

partim:
But those RIPE resources won’t appear under other RIRs, so filtering isn’t necessary?

Anything published by that layer itself will be gone, anyway, under the new rules.

timbru:
indeed! makes sense

partim:
I’ve now added an exception for /0 prefixes and AS ranges covering the full set. I didn’t add an exception for IP ranges because that was a bit complicated due to the way we deal with IP addresses. This should be fine since RFC 3779 says that anything that can be a prefix must be a prefix.

partim:
This PR also introduces a set of new metrics and therefore: Fixes #370.

PR #381, opened 2020-09-22
Support setting the HTTP User Agent in the RRDP client

ximon18:
In `lib.rs` it says:

> "In addition, this also lets you use Routinator as a library for your own dedicated RPKI validation needs"

However, derived clients will still advertise themselves as Routinator/x.y.z when making HTTP RRDP requests which will cause confusion when the server request logs are analyzed if the client was not in fact Routinator but somethingbuilt with the Routinator library.

This patch adds a `set_user_agent(&str)` method to the RRDP HTTP client that can be called after `new()` but before `ignite()` to override the default Routinator/x.y.z HTTP User Agent.

### DIFF
--- a/src/rrdp/http.rs
+++ b/src/rrdp/http.rs
@@ -94,6 +94,25 @@ impl HttpClient {
         })
     }
 
+    pub fn set_user_agent(&mut self, user_agent: &str) -> Result<Self, Error> {
+        match self.client.as_mut() {
+            Ok(_) => {
+                error!("HTTP client is already initialized.");
+                return Err(Error)
+            },
+            Err(builder) => match builder.take() {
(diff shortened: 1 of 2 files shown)


src/rrdp/http.rs
```diff
@@ -55,7 +54,7 @@ impl HttpClient {
 
     pub fn new(config: &Config) -> Result<Self, Error> {
         let mut builder = Client::builder();
-        builder = builder.user_agent(concat!("Routinator/", crate_version!()));
+        builder = builder.user_agent(config.rrdp_user_agent.clone());
```
partim, line 57: You don’t need to clone here. It’ll be happy with a reference.

PR #412, opened 2020-11-01
Fix unknown objects config file setting

johannesmoos:
Rename config file key name ```unknown_objects``` to ```unknown-objects``` (as documentation states)


johannesmoos:
Whoops, looks like I forked from ```master``` instead of ```main```

PR #413, opened 2020-11-01
Fix "unknown-objects" config file key name (typo)

johannesmoos:
Rename config file key name ```unknown_objects``` to ```unknown-objects``` (as documentation states)


partim:
Sorry about the mismatch and thank you for the PR!

Since renaming configuration keys is a breaking change, I would like to allow both keys for now and remove the old one with 0.9. Do you mind if I make a new PR for that? Allowing both is a bit fiddly at that place.

johannesmoos:
Sure, go ahead!

partim:
The PR has been incorporated in #416.

PR #415, opened 2020-11-03
For Issue#414 - handle roa=/32 and prefix < /32.

morrowc:
For cases where the ROA is host-route and a .0 host, if the prefix to compare
is a network at the same .0 the current code fails to understand
validity rules.

ie:
  218.188.106.0/32 - ROA
  218.188.106.0/24 - network to compare.

### DIFF
--- a/src/origins.rs
+++ b/src/origins.rs
@@ -1541,7 +1541,7 @@ pub mod tests {
         assert!(outer.covers(inner_low));
         assert!(outer.covers(inner_mid));
         assert!(outer.covers(inner_hi));
-        assert!(host_roa.covers(outer);
+        assert!(!host_roa.covers(outer);
     }
 }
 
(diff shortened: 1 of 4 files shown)


partim:
Thanks for the fix!

I think we should extend the test to all combinations of prefixes just sure (and rename it to `address_prefix_covers`, since, technically, it tests more than what the name says now). But I am happy to do that after merging the PR if you prefer.

morrowc:
I should note that the change put forth here is the work of:
  https://github.com/aaronw112358

and myself :) we're real live rust programmers now !:) ha!

For the suggested changes, I'll rename and add some more tests in the next short while :)

morrowc:
Ok, some tests and renumbering done.
Let me know how that smells :)

partim:
Ah, darn: misunderstanding. I meant renaming the test not the actual function; that should keep its short name. Sorry!

The tests itself look great!

morrowc:
ha! ok no worries. easy to flip back the function name... I'll tackle that shortly.

partim:
Perfect! Thank you again for the fix and for bearing with me!

PR #425, opened 2020-11-18
Always mark RRDP servers as broken after failed update.

partim:
This will cause falling back to rsync if accessing the repository via RRDP fails.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,9 @@ Breaking Changes
 
 New
 
+* Routinator will now attempt to update repositories via rsync whenever an
+  update via RRDP failed. ([#424])
+
 Bug Fixes
 
 * The prefix validation option incorrectly matched VRPs for host prefixes
@@ -23,6 +26,7 @@ Other Changes
 [#413]: https://github.com/NLnetLabs/routinator/pull/413
 [#415]: https://github.com/NLnetLabs/routinator/pull/415
 [#416]: https://github.com/NLnetLabs/routinator/pull/416
+[#424]: https://github.com/NLnetLabs/routinator/pull/424
 [@johannesmoos]: https://github.com/johannesmoos
 [@morrowc]: https://github.com/morrowc
 [@aaronw112358]: https://github.com/aaronw112358
--- a/src/rrdp/server.rs
+++ b/src/rrdp/server.rs
@@ -133,8 +133,12 @@ impl Server {
         }
 
         let start_time = SystemTime::now();
-        if self.try_update(http, &mut metrics).is_err() && self.check_broken() {
-            let _ = fs::remove_dir_all(self.server_dir.base());
+        if self.try_update(http, &mut metrics).is_err() {
+            if self.check_broken() {
+                let _ = fs::remove_dir_all(self.server_dir.base());
+            }
+            // Mark as broken so we fall back to rsync.
+            self.broken.store(true, Relaxed);
         }
         self.updated.store(true, Relaxed);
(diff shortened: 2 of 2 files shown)


ties:
Hi,

Since I expect the moment at which RP instances detect an RRDP repo to fail are correlated, I am concerned with many RP instances contacting an rsync repository at the same time (thundering herd problem).

When is the next update from rsync scheduled after RRDP fails?

lolepezy:
To continue on the question from @ties: is it possible to have some sort of random delay between routinator notices that RRDP is unavailable and actually uses rsync? It would make the whole process much smoother for the repository side. We are not sure we can/want deploy infrastructure that would reliably handle 2K rsync connections downloading 60K of objects within several minutes window, so it would be nice to spread this load somehow.

ties:
> To continue on the question from @ties: is it possible to have some sort of random delay between routinator noticing that RRDP is unavailable and actually using rsync? It would make the whole process much smoother for the repository side. We are not sure we can/want deploy infrastructure that would reliably handle 2K rsync connections downloading 60K of objects within several minutes window, so it would be nice to spread this load somehow.

My proposal would be to wait for "half the rsync interval + random value from a uniform distribution from 0..."rsync interval". That way the average instance updates one "rsync interval" later while spreading out the load as much as possible.

partim:
Actually, there’s already quite a bit of randomness built in. Routinator starts a new fetch `refresh` seconds after it finished the last one. So the time depends on both when it was started exactly and the overall duration of each consecutive validation run. With `refresh` defaulting to 10 minutes (and the majority of installation apparently sticking to it), I would expect a fairly even distribution of the clients first hitting rsync over a ten minute interval.

One problem I now realize is that if rsync also fails (for instance because the rsync service is has a connection limit), currently Routinator will _not_ use the RRDP data anymore. If it never used rsync, it will not have any data at all.

Maybe doing this whole fallback in such an ad hoc fashion is a bad idea. It is probably better to leave it for 0.9 and the planned redesign of fetching after all.

ties:
> Actually, there’s already quite a bit of randomness built in. Routinator starts a new fetch `refresh` seconds after it finished the last one. So the time depends on both when it was started exactly and the overall duration of each consecutive validation run. With `refresh` defaulting to 10 minutes (and the majority of installation apparently sticking to it), I would expect a fairly even distribution of the clients first hitting rsync over a ten minute interval.

I will need to do a more thorough analysis, but my estimate is that spreading the clients out over the rsync update interval would really have a big impact on the load on rsync servers, it is a 6x reduction in load.

Do you think routinator will keep the rsync directories when rrdp works again? If so I will try to get some data on the duration of rsync runs for both completely expired, relatively recent, and empty rsync directories.

partim:
> I will need to do a more thorough analysis, but my estimate is that spreading the clients out over the rsync update interval would really have a big impact on the load on rsync servers, it is a 6x reduction in load.

The requests should basically be spread exactly the same way as the RRDP requests are right now. Admittedly I haven’t checked how that actually looks, but that should be fairly easy.

> Do you think routinator will keep the rsync directories when rrdp works again?

I believe it currently does – I think there isn’t actually any cleanup done.

> If so I will try to get some data on the duration of rsync runs for both completely expired, relatively recent, and empty rsync directories.

I’d love to hear your thoughts on operational considerations for as graceful a fallback as possible.

Given that there might be some unintended consequences from just falling back in the current architecture, I’ll will close this PR for now. We will properly implement fallback as part of the new architecture.

PR #431, opened 2020-11-23
Add package repo instructions to readme

AlexanderBand:
Maybe it needs a reference to the location of the config file as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,11 +80,14 @@ used in Cargo.
 sudo apt install routinator
 sudo routinator-init
 # Follow instructions provided
-systemctl enable --now routinator
+sudo systemctl enable --now routinator
 ```
 
-You can check the status of Routinator with `sudo systemctl status 
-routinator` and view the logs with `sudo journalctl --unit=routinator`.
+By default, Routinator will start the RTR server on port 3323 and the HTTP
+server on port 8323. These, and other values can be changed in the
+configuration file located in `/etc/routinator/routinator.conf`. You can check
+the status of Routinator with `sudo systemctl status  routinator` and view the
+logs with `sudo journalctl --unit=routinator`.
 
 ## Quick Start with Docker
 
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ sudo apt update
 
 You can then install, initialise, enable and start Routinator by running these
 commands. Note that `routinator-init` is slightly different than the command
-used in Cargo.
+used with Cargo.
 
 ```bash
 sudo apt install routinator
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ sure to install the new TALs using
 routinator init --force
 ```
 
-## Quick Start with Pre-built Packages
+## Quick Start with Debian and Ubuntu Packages
 
 ### Disclaimer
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,6 +46,44 @@ sure to install the new TALs using
 routinator init --force
 ```
 
+## Quick Start with Pre-built Packages
```
partim, line 49: Perhaps instead of ‘Pre-built Packages’ specifically saying ’Debian and Ubuntu’ would be better?

README.md
```diff
@@ -46,6 +46,44 @@ sure to install the new TALs using
...
+```
+Then run the following commands.
+
+```bash
+sudo apt update && apt-get install -y gnupg2
+wget -qO- https://packages.nlnetlabs.nl/aptkey.asc | sudo apt-key add -
+sudo apt update
+```
+
+You can then install, enable and start Krill by running
```
partim, line 75: s/Krill/Routinator/

README.md
```diff
@@ -46,6 +46,44 @@ sure to install the new TALs using
...
+
+```bash
+sudo apt install routinator
+sudo routinator-init --accept-arin-rpa
```
partim, line 79: So far I always avoided providing a pastable snippet with the `--accept-arin-rpa` in it so that people are forced to at least consider reading the RPA. I think this would be good here, too.

ximon18, line 79: Is it worth calling out in some way that this is **NOT** the usual `routinator init` command (but looks very much like it) ?... maybe it wasn't such a great idea naming it like that... not sure.

README.md
```diff
@@ -46,6 +46,48 @@ sure to install the new TALs using
...
+used in Cargo.
+
+```bash
+sudo apt install routinator
+sudo routinator-init
+# Follow instructions provided
+systemctl enable --now routinator
+```
+
+routinator-init
```
ximon18, line 86: Sorry, I don't get it, what is this? `routinator-init` is arleady mentioned above.

AlexanderBand: Oops

README.md
```diff
@@ -46,6 +46,48 @@ sure to install the new TALs using
...
+
+You can then install, initialise, enable and start Routinator by running these
+commands. Note that `routinator-init` is slightly different than the command
+used in Cargo.
+
+```bash
+sudo apt install routinator
+sudo routinator-init
+# Follow instructions provided
+systemctl enable --now routinator
```
ximon18, line 83: This needs `sudo` too.

partim:
> Maybe it needs a reference to the location of the config file as well?

That would be good, indeed. Maybe also mention the ports it’ll listen on if you use the default config file?

README.md
```diff
@@ -46,6 +46,49 @@ sure to install the new TALs using
...
+
+```bash
+sudo apt update && apt-get install -y gnupg2
+wget -qO- https://packages.nlnetlabs.nl/aptkey.asc | sudo apt-key add -
+sudo apt update
+```
+
+You can then install, initialise, enable and start Routinator by running these
+commands. Note that `routinator-init` is slightly different than the command
+used in Cargo.
```
ximon18, line 77: used with Cargo.

ximon18 (approved):

PR #445, opened 2021-01-08
check if /var/lib/routinator directory exists before starting routinator-init

nickhilliard:
On a newly installed system, the `debian/routinator-init` script fails because /var/lib/routinator does not exist:

```
rpki01:/root# routinator-init
Fatal: failed to create cache directory /var/lib/routinator/rpki-cache: Permission denied (os error 13)
rpki01:/root#
```

this pull requests tests to ensure that the `/var/lib/routinator` directory exists, and if it doesn't, then it creates it + changes ownership.

### DIFF
--- a/debian/routinator-init
+++ b/debian/routinator-init
@@ -1,2 +1,8 @@
 #!/bin/sh
+
+if [ ! -d /var/lib/routinator ]; then
+        mkdir /var/lib/routinator
+        chown routinator:routinator /var/lib/routinator
+fi
+
 sudo -u routinator routinator --config /etc/routinator/routinator.conf init "$@"


ximon18:
Thanks @nickhilliard.

Before applying your patch I'd like to see if I can understand what is going on.

Which operating system and version are you using?

I ask because we have automated tests for several flavours of Debian and Ubuntu which install the package and during installation it creates the directory that is missing in your case, e.g. from [this test run](https://github.com/NLnetLabs/routinator/runs/1656379358?check_suite_focus=true#step:10:68):

```
Setting up routinator (0.9.0~dev-1xenial) ...
Adding system user `routinator' (UID 107) ...
Adding new group `routinator' (GID 112) ...
Adding new user `routinator' (UID 107) with group `routinator' ...
Creating home directory `/var/lib/routinator/' ...
routinator.service is a disabled or a static unit, not starting it.
```

Following that the tests successfully invoke the `routinator-init` script, e.g. from [here](https://github.com/NLnetLabs/routinator/runs/1656379358?check_suite_focus=true#step:11:99):

```
INIT ROUTINATOR:
Created local repository directory /var/lib/routinator/rpki-cache
Installed 5 TALs in /var/lib/routinator/tals
+ echo -e '\nROUTINATOR DATA DIR AFTER INIT:'

ROUTINATOR DATA DIR AFTER INIT:
+ sg lxd -c 'lxc exec testcon -- ls -la /var/lib/routinator'
total 16
drwxr-xr-x  4 routinator routinator 4096 Jan  6 12:46 .
drwxr-xr-x 28 root       root       4096 Jan  6 12:46 ..
drwxr-xr-x  2 routinator routinator 4096 Jan  6 12:46 rpki-cache
drwxr-xr-x  2 routinator routinator 4096 Jan  6 12:46 tals
```

So I'm wondering what is different on your system? Do you see anything different in our tests compared to the commands that were executed on your system? You can see the commands issued by the test [here](https://github.com/NLnetLabs/routinator/blob/main/.github/workflows/pkg.yml).

Thanks,

Ximon

nickhilliard:
ok, so the routinator user already exists on my system with a different home directory (excuse for brazen lie: am an end-user), which means that it's not recreating that user, which means that /var/lib/routinator doesn't exist.  So not a bug on the default system.  Sorry for the noise.

ximon18:
No problem @nickhilliard, please don’t hesitate to contact us again if needed.

PR #447, opened 2021-01-25
Manually set std listeners sockets to non-blocking.

partim:
This is necessary because `tokio::net::TcpListener::from_std` won’t do it and so we block.

(This is a bug fix for not-yet-released code; no changelog entry necessary.)


DRiKE (approved):
Looks good to me!

PR #470, opened 2021-02-19
Add --fresh to wipe all local data before starting.

partim:
This option can be used in case of a database corruption.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,12 @@ Breaking Changes
 
 New
 
+* Routinator now keeps the last valid data from a publication point and
+  falls back to using that if an update to the publication point does not
+  have a valid manifest or the data does not match the manifest. ([#456]).
+* The new option `--fresh` causes Routinator to delete all locally stored
+  data before starting. This can be used when data corruption is
+  reported. ([#470])
 * Status information is now available in JSON format at `/api/v1/status`.
   ([#437])
 * The RRDP client now supports the gzip transfer encoding for HTTPs.
@@ -26,7 +32,9 @@ Other Changes
 [#437]: https://github.com/NLnetLabs/routinator/pull/437
 [#443]: https://github.com/NLnetLabs/routinator/pull/443
 [#444]: https://github.com/NLnetLabs/routinator/pull/444
+[#456]: https://github.com/NLnetLabs/routinator/pull/456
 [#463]: https://github.com/NLnetLabs/routinator/pull/463
+[#470]: https://github.com/NLnetLabs/routinator/pull/470
 [rpki-rs]: https://github.com/NLnetLabs/rpki-rs/
 [rpki-rtr]: https://github.com/NLnetLabs/rpki-rtr/
 [@bjpbakker]: https://github.com/bjpbakker
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,9 +12,8 @@ New
 * Routinator now keeps the last valid data from a publication point and
   falls back to using that if an update to the publication point does not
   have a valid manifest or the data does not match the manifest. ([#456]).
-* The new option `--fresh` causes Routinator to delete all locally stored
-  data before starting. This can be used when data corruption is
-  reported. ([#470])
+* The new option `--fresh` causes Routinator to delete all cached data
+  before starting. This can be used when data corruption is reported. ([#470])
 * Status information is now available in JSON format at `/api/v1/status`.
   ([#437])
 * The RRDP client now supports the gzip transfer encoding for HTTPs.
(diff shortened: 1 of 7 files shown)


Changelog.md
```diff
@@ -9,6 +9,12 @@ Breaking Changes
 
 New
 
+* Routinator now keeps the last valid data from a publication point and
+  falls back to using that if an update to the publication point does not
+  have a valid manifest or the data does not match the manifest. ([#456]).
+* The new option `--fresh` causes Routinator to delete all locally stored
+  data before starting. This can be used when data corruption is
+  reported. ([#470])
```
ximon18, line 17: Perhaps refine this to say this can be used when _Routinator_ reports that its data is corrupted? Does Routinator actually report that?

Changelog.md
```diff
@@ -26,7 +32,9 @@ Other Changes
 [#437]: https://github.com/NLnetLabs/routinator/pull/437
 [#443]: https://github.com/NLnetLabs/routinator/pull/443
 [#444]: https://github.com/NLnetLabs/routinator/pull/444
+[#456]: https://github.com/NLnetLabs/routinator/pull/456
```
ximon18, line 35: This line is unrelated to this PR as far as I can see, it was just missed from PR #456. If the intention is to strictly relate changes to the issues they address then perhaps this and the reference above to 456 should be pulled out of this PR and pushed separately?

Changelog.md
partim: It does. And then exits.

I left out ‘Routinator’ there because of word repetition.

doc/routinator.1
```diff
@@ -240,6 +240,12 @@ explicit port.
 .IP
 This option allows to disable this filtering.
 
+.TP
+.B --fresh
+Delete and re-initialize all local data storage before starting. This option
```
ximon18, line 245: "All", but not including for example TAL files, does that need to be clarified?

src/config.rs
```diff
@@ -135,6 +135,11 @@ pub struct Config {
     /// Allow dubious host names.
     pub allow_dubious_hosts: bool,
 
+    /// Should we wipe store and cache before starting?
```
ximon18, line 138: The config argument documentation refers to "store", not "cache", the code changes seem only to affect "caches" and this comment refers to both "store" and "cache". Would it be helpful to be more consistent in the naming?

src/config.rs
```diff
@@ -311,6 +316,10 @@ impl Config {
              .long("allow-dubious-hosts")
              .help("Allow dubious host names in rsync and HTTPS URIs")
         )
+        .arg(Arg::with_name("fresh")
+            .long("fresh")
+            .help("Delete all locally stored data, download everything again")
```
ximon18, line 321: Stupid question: is it possible to pass `--fresh` to an invocation of Routinator that will NOT download anything thus invalidating this description?

src/process.rs
```diff
@@ -200,7 +200,8 @@ impl Process {
                 .level_for("tokio_reactor", LevelFilter::Info)
                 .level_for("hyper", LevelFilter::Info)
                 .level_for("reqwest", LevelFilter::Info)
-                .level_for("h2", LevelFilter::Info);
+                .level_for("h2", LevelFilter::Info)
```
ximon18, line 203: These seem like unrelated changes, do they relate to #456? Do they need mentioning in the changelog? Should they be separated out from this PR?

ximon18 (commented):
LGTM

Changelog.md
partim: I don’t feel like we need to be too concerned about the changelog. As long as it has everything noteworthy upon release.

ximon18 (approved):

doc/routinator.1
partim: The term ‘local data storage’ was meant to convey all the data that Routinator itself is creating.

It now occurs to me that ‘cache dir’ (as used by configuration) is incorrect now. Let’s clean that up with a separate PR. (The whole thing should be called the cache and what is now `routinator::cache` should be renamed into something else.

src/config.rs
partim: Yes. This might be worthwhile mentioning in the man page. The help text needs to be short.

src/process.rs
partim: Given that sled was only introduced in #456, I don’t think we need to mention changes made to changes initially described in the changelog for the same release.

This should probably have been a separate PR but it is just a one liner.

partim:
I have adjusted documentation to refer to all data created by Routinator as ‘cache.’ What is currently referred to as the cache in the source, i.e., `routinator::cache`, will be renamed in a separate PR.

PR #471, opened 2021-02-19
Also accept a string for the exceptions config file value.

### LINKED ISSUE #464: Accept both a string or an array of strings for SLURM files
When pointing to a SLURM file in the config file, Routinator requires an array of strings. Please also accept just a string.

partim:
Fixes #464.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,7 +14,7 @@ New
 * The RRDP client now supports the gzip transfer encoding for HTTPs.
   ([#463], contributed by [@bjpbakker])
 * The `exception` config file value now also accepts a single string with
-  a path name in addition to an array of strings. ([#471])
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -13,6 +13,8 @@ New
   ([#437])
 * The RRDP client now supports the gzip transfer encoding for HTTPs.
   ([#463], contributed by [@bjpbakker])
+* The `exception` config file value now also accepts a single string with
+  a path name in addition to an array of strings. ([#471])
```
ximon18, line 17: "in addition to" makes it sound like you can pass something like `exception = "/some/path", [ "some", "array", "values" ]` which I guess is not the case. Perhaps "instead of" rather than "in addition to" ?

ximon18 (approved):
LGTM

PR #474, opened 2021-02-24
 Send Last-Modified response header field with validation result (fixes #469)

### LINKED ISSUE #469: Feature Request: HTTP serve validation results with Last-Modified header field
I'm currently running tests with various validators, GETting JSON output and post-processing it. It would be very useful if the JSON output would be served with a Last-Modified header field, indicating the time when the last validation run finished.

(see also https://github.com/cloudflare/cfrpki/issues/83)

reschke:


src/http.rs
```diff
@@ -966,6 +967,9 @@ fn vrps(
         }
     };
 
+    let (_start, done, _duration) = origins.update_times();
+    let http_date = StrftimeItems::new("%a, %d %h %Y %H:%M:%S GMT");
```
partim, line 971: Ideally, this would be a constant with an array of `chrono::format::Item`s. In its current form, the string needs to be parsed for each request which is a bit wasteful since it never changes.

src/http.rs
```diff
@@ -977,6 +981,7 @@ fn vrps(
     Response::builder()
     .header("Content-Type", format.content_type())
     .header("content-length", stream.output_len())
+    .header("Last-Modified", done.unwrap().format_with_items(http_date.clone()).to_string())
```
partim, line 984: While unwrapping here should be fine here right now, it is probably still better to only add the header when it is actually available. Maybe you can wrap it in an `if let Some(done) = done { }` block somehow?

partim (changes requested):
Thank you for the pull request! I have two comments inline.

src/http.rs
reschke: Thanks for the feedback.

The issue is: this is the first time ever I'm trying to code in Rust, and I have to say that the documentation is not simple to understand.

I'll try this though.

src/http.rs
reschke: I had a look at the CSV date handling, but couldn't find how to format the weekday as string.

Naively (perhaps), I would hope that I can use the format string to create the Items just once, but I didn't get that to work with "const". As I said, no experience with Rust.

src/http.rs
partim: You probably have to break it up into three parts using variables.

You could either use a mutable variable and reassign to it:

```
let mut builder = Response::builder();
if let Some(done) = done {
    builder = builder.header( /* ... */);
}
builder.more_stuff()
```

Or, the move the value along to a new variable (re-using the same name which is fine):

```
let builder = Builder::builder();
let builder = if let Some(done) = done {
    builder.header( /* ... */)
} else {
    builder
};
builder.more_stuff()
```

The latter is probably more proper but also more wordy, so I would probably go for the former.

src/http.rs
partim: No worries!

Looks like the week day and month names are called ‘fixed’. So, the week day would be a `Item::Fixed(Fixed::ShortWeekdayName)` (assuming you import the bare names by doing a `use chrono::format::{Item, Fixed}` up top).

The string parsing is just a regular function so it has to happen at runtime. (Rust has something called `const fn` as functions that are evaluated by the compiler but they are somewhat limited in what they can do.)

partim:
> but if we weren't done, would we get here in the first place?

No, but this way the code is more robust. The call to `unwrap` essentially makes the program crash if the option is `None` which can happen if `origin.update_times` is being changed and the unwrap here is overlooked during refactoring.

src/http.rs
reschke: Almost there. Thanks for the help so far. In Java, for instance, I would be able to build a static final DateFormat thingy by parsing the format string once.

Anyway, is some magic needed to ensure that this will always use the ENGLISH locale?

reschke:
> The call to unwrap essentially makes the program crash if the option is None which can happen if origin.update_times is being changed and the unwrap here is overlooked during refactoring.

Ack.

src/http.rs
```diff
@@ -974,10 +994,16 @@ fn vrps(
         current, filters, metrics
     );
 
-    Response::builder()
-    .header("Content-Type", format.content_type())
-    .header("content-length", stream.output_len())
-    .body(Body::wrap_stream(stream::iter(
+    let mut builder = Response::builder();
+
+    builder = builder.header("Content-Type", format.content_type())
```
partim, line 999: You don’t need a new assignment here, this can all be done in one fell swoop.

(As a side note: Can you make sure the line length is not more than 78 characters? I need to document that somewhere as it has become nonstandard by now …)

src/http.rs
partim: If I understand the chrono crate correctly, enabling locales requires adding a feature. So we should be fine.

It might be worthwhile to add a test to make sure?

reschke: For that I'd need much more hand holding :-)

partim: I can also do that later, unless you want to give it a spin?

reschke: "Later" sounds good to me...

partim (approved):

partim:
Once again, thank you for the PR! I hope you enjoyed your venture into Rust land a little and it wasn’t only frustrating.

reschke:
Thanks a lot for the help.

Rust is interesting, but it's harder to get into than I expected :-)

PR #480, opened 2021-03-15
RRDP Collector: Always store the repository state information under the same key. 

partim:
This fixes a bug where delta updates where re-tried again after being applied and then failed.


ximon18 (approved):
LGTM!

PR #486, opened 2021-03-18
Clear validation log after cleanup is done.

partim:
Currently, the cleanup log shows in `/log` as the first thing of the next validation run. And that is wrong.


DRiKE (approved):
Looks good.

density215 (approved):

Issue #477, opened 2021-03-11
Feature: RRDP serial as a metric


ties:
As a repository operator, to be sure that my repository is getting updates, I like to monitor that the RRDP serial is increasing.  For this I would like to have the RRDP serial as a Prometheus metric.

```yaml
name: SerialReset
expr: rrdp_serial{address=~".*[host]/.*"} < max_over_time(rrdp_serial{address=~".*[host]/.*"}[30m])
labels:
severity: P4
annotations:
description: {{ $labels.group }} RRDP serial at {{ $labels.address }} reset from {{ with printf "max_over_time(rrdp_serial{address='%s'}[30m])" .Labels.address | query }}{{ . | first | value }}{{end}} to {{ $value }} (in the last 30m) according to {{ $labels.instance }}.
summary: {{ $labels.group }} RRDP serial at {{ $labels.address }} reset to {{ $value }}.
```

For bonus points, track accepted/rejected rrdp updates as well, such as OctoRPKI does:
```yaml
name: OctoRPKIRRDPErrors
expr: rrdp_errors{address="https://[host]/rrdp/notification.xml"} > 0
labels:
severity: P1
annotations:
description: OctoRPKI found {{ humanize $value }} RRDP errors for {{ $labels.address} from {{ $labels.instance }}.
summary: OctoRPKI found {{ humanize $value }} RRDP errors for {{ $labels.address}} on {{ $labels.hostname }}
```

```yaml
name: OctoRPKIRRDPDeltaIncreasing
expr: (increase(rrdp_serial{address=~".*[host]/.*"}[1h30m]) == 0)
labels:
severity: P1
annotations:
description: OctoRPKI did see {{ $value }} new RRDP serials at {{ $labels.address }} during the last 90  minutes from {{ $labels.instance }}.
summary: No new RRDP serial at {{ $labels.address }} for 90m.
```

PR #487, opened 2021-03-18
Add additional status information to RRDP repository metrics.

partim:
This partially fixes issue #477.


src/http.rs
```diff
@@ -662,16 +679,17 @@ fn status_active(
             }).unwrap_or(-1),
         ).unwrap();
         if let Ok(duration) = metrics.duration {
-            writeln!(
+            write!(
                 res,
                 ", duration={:.3}s",
                 duration.as_secs() as f64
```
DRiKE, line 685: maybe using `as_secs_f64()` can make this more concise

DRiKE (commented):
typo and small nit only, looks good otherwise

src/http.rs
partim: Uh, didn’t even realize that existed.

PR #488, opened 2021-03-23
Conditional processing for VRP endpoints of the HTTP server.

### LINKED ISSUE #469: Feature Request: HTTP serve validation results with Last-Modified header field

partim:
This PR adds adding of Etag headers to responses (with Last-Modfied already added previously) and processes If-None-Match and If-Modified-Since headers in requests to all endpoints of the HTTP server providing a list of VRPs.

Resolves #469.

@reschke, if you have the time, maybe you can have a quick look over the implementation of request handling in `src/http.rs`’s `maybe_not_modified` function to check that we follow good HTTP practice?

### DIFF
--- a/src/http.rs
+++ b/src/http.rs
@@ -16,13 +16,14 @@ use std::pin::Pin;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::task::{Context, Poll};
-use chrono::{Duration, Utc};
+use chrono::{DateTime, Duration, Utc};
 use chrono::format::{Item, Fixed, Numeric, Pad};
 use clap::{crate_name, crate_version};
 use futures::stream;
 use futures::pin_mut;
 use futures::future::{pending, select_all};
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
+use hyper::header::HeaderValue;
 use hyper::server::accept::Accept;
 use hyper::service::{make_service_fn, service_fn};
 use log::error;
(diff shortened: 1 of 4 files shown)


reschke (commented):
Hm, I'm still bad at reading Rust code.

Two things you may want to check:

- edge case: splitting conditional header fields on "," can fail if the ETag happens to contain a "," (but you might not care as your etags do not)
- matching if-modified-since needs to handle the case where the ts changed within the same 1sec interval (due to the http-date resolution)

I can try to build this and run some blackbox tests though...

partim:
> edge case: splitting conditional header fields on ","

Hm, I guess there could be a case where a legit etag follows later. But then, if this fails, the full content is served, so this fails to the safe side.

> matching if-modified-since needs to handle the case where the ts changed within the same 1sec interval

Yikes. It’s fairly unlikely to happen as real-world validation runs take longer than a second, but I guess we should handle this anyway. This may be something that breaks test infrastructure in fun ways.

partim:
I think this is now handling those two cases properly¹. Thanks for the review!

___
¹ In the case of the etag parsing, it will accept a correct tag followed by garbage, which I think is fine.

reschke:
> In the case of etag parsing, it will accept a correct tag followed by garbage, which I think is fine.

Absolutely; after all, this is a special purpose server. If a client insists on sending ETag not generated by the server, so be it :-)

src/http.rs
```diff
@@ -1021,55 +1002,80 @@ fn vrps(
...
+/// `None`.
+fn maybe_not_modified(
+    req: &Request<Body>,
+    etag: &str,
+    done: DateTime<Utc>,
+) -> Option<Response<Body>> {
+    // First, check If-None-Match.
+    for value in req.headers().get_all("If-None-Match").iter() {
+        // Skip ill-formatted values. By being lazy here we may falsely
+        // return a full response, so this should fine.
```
DRiKE, line 1044:
```suggestion
        // return a full response, so this should be fine.
```

src/http.rs
```diff
@@ -1232,6 +1265,156 @@ impl Drop for HttpStream {
...
+/// makes this indistinguishable from reaching the end of a correctly
+/// formatted value. As a consequence, we will 304 a request that has the
+/// right tag followed by garbage.
+struct EtagsIter<'a>(&'a str);
+
+impl<'a> Iterator for EtagsIter<'a> {
+    type Item = &'a str;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Skip white space and check with if we are done.
```
DRiKE, line 1284:
```suggestion
        // Skip white space and check if we are done.
```

src/http.rs
```diff
@@ -1232,6 +1265,156 @@ impl Drop for HttpStream {
...
+        }
+
+        Some(res)
+    }
+}
+
+
+//------------ Parsing and Constructing HTTP Dates ---------------------------
+
+/// Definition of the preferred date format (aka IMF-fixedate).
```
DRiKE, line 1324:
```suggestion
/// Definition of the preferred date format (aka IMF-fixdate).
```

src/http.rs
```diff
@@ -1232,6 +1265,156 @@ impl Drop for HttpStream {
...
+///
+/// The definition allows for relaxed parsing: It accepts additional white
+/// space and ignored case for textual representations. It does, however,
```
DRiKE, line 1327:
```suggestion
/// space and ignores case for textual representations. It does, however,
```

DRiKE (approved):
Small typos only, see suggestions

PR #489, opened 2021-03-23
Further improve RRDP metrics.

### LINKED ISSUE #477: Feature: RRDP serial as a metric

partim:
This PR further enhances the RRDP repository metrics and resolves #477.

This changes the meaning of ‘status’ to be the status code of the last HTTP request made with separate status information available for the notification file and the series of payload requests. In addition, if a snapshot was requested, the reason for that is given.

@ties, is it enough to have the new combined status in the Prometheus or do you rather have more Prometheus metrics and if so which?

### DIFF
--- a/src/http.rs
+++ b/src/http.rs
@@ -356,8 +356,8 @@ fn metrics_active(
     // rrdp_status
     writeln!(res,
         "\n\
-        # HELP routinator_rrdp_status status code for getting \
-            notification file\n\
+        # HELP routinator_rrdp_status combined status code for repository \
+            update requests\n\
         # TYPE routinator_rrdp_status gauge"
     ).unwrap();
     for metrics in metrics.rrdp() {
@@ -366,6 +366,42 @@ fn metrics_active(
             "routinator_rrdp_status{{uri=\"{}\"}} {}",
             metrics.notify_uri,
             metrics.status().map(|code| {
+                code.as_u16() as i16
+            }).unwrap_or(-1),
+        ).unwrap();
+    }
+
+    // rrdp_notification_status
+    writeln!(res,
+        "\n\
+        # HELP routinator_rrdp_notification_status status code for getting \
+            notification file\n\
+        # TYPE routinator_rrdp_notification_status gauge"
+    ).unwrap();
+    for metrics in metrics.rrdp() {
+        writeln!(
+            res,
+            "routinator_rrdp_notification_status{{uri=\"{}\"}} {}",
+            metrics.notify_uri,
+            metrics.notify_status.map(|code| {
+                code.as_u16() as i16
+            }).unwrap_or(-1),
+        ).unwrap();
+    }
+
+    // rrdp_payload_status
+    writeln!(res,
+        "\n\
+        # HELP routinator_rrdp_payload_status status code(s) for getting \
+            payload file(s)\n\
+        # TYPE routinator_rrdp_payload_status gauge"
+    ).unwrap();
+    for metrics in metrics.rrdp() {
+        writeln!(
+            res,
+            "routinator_rrdp_payload_status{{uri=\"{}\"}} {}",
+            metrics.notify_uri,
+            metrics.payload_status.map(|code| {
                 code.as_u16() as i16
             }).unwrap_or(-1),
         ).unwrap();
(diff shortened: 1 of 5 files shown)


ties:
Hi,

This is a nice addition for my use case.

I tried to build this branch (clone, `cargo build`, `./target/debug/routinator -c ~/.rpki-cache/config -vv server`) but I don't see some of the metrics that I would expect based on the changes and the description.

However I do see that the `routinator_rrdp_status` metric sometimes turns to 1 when there is no failed request in the logs. What does this value mean?
```
# HELP routinator_rrdp_status status code for getting notification file
# TYPE routinator_rrdp_status gauge
routinator_rrdp_status{uri="https://rrdp.ripe.net/notification.xml"} -1
...

# HELP routinator_rrdp_serial serial number of last RRDP update
# TYPE routinator_rrdp_serial gauge
routinator_rrdp_serial{uri="https://rrdp.ripe.net/notification.xml"} 669
```
The rrdp serial is perfect for my use case (detecting liveliness of publication points) :).

partim:
The missing metrics aren’t in the Prometheus metrics but only in the /status and /api/v1/status endpoints. I probably could add at least notify-status and payload-status there as well, if that is useful.

*  *  *
A status code of -1 is supposed to mean that fetching the response failed, so probably an IO error. There should indeed be log output further explaining the issue. There might be a mistake in how the value is derived from notify-status and  payload-status. I’ll have a look.

partim:
Ah! You get the -1 payload-status code when there was no payload request because the data was up-to-date. That seems suboptimal. I’ll figure something out.

partim:
This should be supported now. The payload status is now 0 if there was no payload query necessary.

ties:
I did not know about `/api/v1/status` :). The prometheus endpoint looks very good now and this definitely covers my use case.

src/http.rs
```diff
@@ -838,9 +881,15 @@ fn api_status(origins: &OriginsHistory) -> Response<Body> {
...
                     target.member_raw("delta",
-                        if metrics.delta { "true" }
+                        if metrics.snapshot_reason.is_none() { "true" }
                         else { "false" }
                     );
+                    if let Some(reason) = metrics.snapshot_reason {
+                        target.member_str("snapshot_reason", reason.code())
+                    }
+                    else {
+                        target.member_raw("snapshot_reason", "none");
```
ties, line 891: I don't think `none` is a valid json value

partim: Argl. Fixed.

partim:
`/api/v1/status` was introduced in 0.8.3 to supply status information to the UI.

density215 (approved):

PR #490, opened 2021-03-24
Optionally store RRDP responses somewhere.

### LINKED ISSUE #476: storing downloaded RRDP objects
As an engineer I want to have all RRDP deltas available when debugging so that I can investigate transient issues.

## Use case
Sometimes I am investigating transient issues that disappear very quickly.
Screenshot 2021-03-04 at 10 13 02

In this case one of the TALs dropped a large ROA and it was back in delta. For occurrences like this is easy if you can drill down to the exact objects. I have used http://www.rpkiviews.org/ for this but the 10 minute granularity is not enough for some situations.

In cases like this it would help me if I could configure routinator to save the deltas

partim:
This PR introduces the `rrdp-keep-responses` option that allows providing a directory in which the content of successfully received RRDP responses is stored. The URIs are turned into a path. For notification files, the URI is turned into a directory and the notification file is stored under the timestamp when it was retrieved.

Resolves #476.

### DIFF
--- a/src/collector/rrdp.rs
+++ b/src/collector/rrdp.rs
@@ -1221,7 +1221,7 @@ impl HttpResponse {
         })
     }
 
-    /// Returns the value of the content length header of present.
+    /// Returns the value of the content length header if present.
     pub fn content_length(&self) -> Option<u64> {
         self.response.content_length()
     }
--- a/src/collector/rrdp.rs
+++ b/src/collector/rrdp.rs
@@ -1189,7 +1189,7 @@ impl io::Read for HttpResponse {
     fn read(&mut self, buf: &mut [u8]) -> Result<usize, io::Error> {
         let res = self.response.read(buf)?;
         if let Some(file) = self.file.as_mut() {
-            file.write(&buf[..res])?;
+            file.write_all(&buf[..res])?;
         }
         Ok(res)
     }
--- a/src/collector/rrdp.rs
+++ b/src/collector/rrdp.rs
@@ -1088,7 +1088,7 @@ impl HttpClient {
     /// Performs an HTTP GET request for the given URI.
     ///
     /// If keeping responses is enabled, the response is written to a file
-    /// corresponding to the URI. If resource behind the URI changes over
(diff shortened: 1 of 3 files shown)


src/collector/rrdp.rs
```diff
@@ -1091,6 +1104,98 @@ impl HttpClient {
...
+    }
+
+    pub fn content_length(&self) -> Option<u64> {
+        self.response.content_length()
+    }
+
+    pub fn copy_to<W: io::Write + ?Sized>(
+        &mut self, w: &mut W
+    ) -> Result<u64, io::Error> {
+        io::copy(self, w)
```
DRiKE, line 1180: Consider adding a comment explaining that the `fn read` in `impl io::Read for HttpResponse` is what actually decides what is being written here because of the Reader/Writer interplay in `io::copy` (i.e. only the `.response` from the HttpResponse struct, not the entire struct including the optional `.file`).

DRiKE (commented):
Some clarification needed in the comments, otherwise good to go.

src/collector/rrdp.rs
```diff
@@ -1170,21 +1181,30 @@ impl HttpResponse {
         })
     }
 
+    /// Returns the value of the content length header of present.
```
DRiKE, line 1224:
```suggestion
    /// Returns the value of the content length header if present.
```

src/collector/rrdp.rs
```diff
@@ -1048,9 +1048,9 @@ impl HttpClient {
     /// Performs an HTTP GET request for the given URI.
     ///
     /// If keeping responses is enabled, the response is written to a file
-    /// corresponding to the URI. If this would not result in a unique file
-    /// name, set `multi` to `true` to also include the current time in
-    /// the name.
+    /// corresponding to the URI. If resource behind the URI changes over
```
DRiKE, line 1091:
```suggestion
    /// corresponding to the URI. If a/the resource behind the URI changes over
```

DRiKE (approved):

PR #493, opened 2021-03-26
Produce per-repository metrics.

### LINKED ISSUE #338: Feature: invalid/rejected objects by repository
As a user operating a repository, I want to be able to monitor the number (or existence) of invalid&rejected objects per repository, so that I can monitor the health of my repository without false positives.

rpki-client gives overall statistics for which [you can expose using a container](https://github.com/ties/rpki-client-web) but this is noisy for alerting since it is per trust anchor.

Ideally I would like stats per repository _or_ the ability to limit the repositories being considered for validation.

rpki-client stats including the lacnic + RIPE tal:
```
# HELP rpki_objects Number of objects by type
# TYPE rpki_objects gauge
rpki_objects{type="roas"} 16356.0
rpki_objects{type="failedroas"} 0.0
rpki_objects{type="invalidroas"} 0.0
rpki_objects{type="certificates"} 11873.0
rpki_objects{type="failcertificates"} 0.0
rpki_objects{type="invalidcertificates"} 0.0
rpki_objects{type="manifests"} 11873.0
rpki_objects{type="failedmanifests"} 3.0
rpki_objects{type="stalemanifests"} 0.0
rpki_objects{type="crls"} 11870.0
rpki_objects{type="repositories"} 13.0
rpki_objects{type="vrps"} 88325.0
rpki_objects{type="uniquevrps"} 88325.0
```

partim:
This PR introduces a whole lot more metrics related to publication and validation. The already existing metrics over VRPs are now also provided per repository. In addition, metrics over the number and status of published objects are available per TAL, per repository, and overall.

Fixes #338.

### DIFF
--- a/src/collector/rrdp.rs
+++ b/src/collector/rrdp.rs
@@ -499,7 +499,7 @@ impl<'a> Run<'a> {
     /// If you are not interested in the metrics, you can simple drop the
     /// value, instead.
     pub fn done(self, metrics: &mut Metrics) {
-        metrics.set_rrdp(self.metrics.into_inner().unwrap())
+        metrics.rrdp = self.metrics.into_inner().unwrap()
     }
 }
 
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -19,13 +19,14 @@
 /// points.
 
 use std::{fmt, fs, io, str};
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::fs::File;
 use std::path::{Path, PathBuf};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::sync::atomic::{AtomicBool, Ordering};
 use bytes::Bytes;
-use crossbeam_queue::SegQueue;
+use crossbeam_queue::{ArrayQueue, SegQueue};
 use crossbeam_utils::thread;
 use log::{debug, error, warn};
 use rpki::repository::cert::{Cert, KeyUsage, ResourceCert};
@@ -41,7 +42,9 @@ use crate::{collector, store};
 use crate::collector::Collector;
 use crate::config::{Config, FilterPolicy};
 use crate::error::Failed;
-use crate::metrics::Metrics;
+use crate::metrics::{
+    Metrics, PublicationMetrics, RepositoryMetrics, TalMetrics
+};
 use crate::origins::OriginsReport;
 use crate::store::{Store, StoredManifest};
 use crate::utils::str_from_ascii;
@@ -337,7 +340,7 @@ impl Engine {
         &self
     ) -> Result<(OriginsReport, Metrics), Failed> {
         let report = OriginsReport::new();
-        let run = self.start(&report)?;
+        let mut run = self.start(&report)?;
         run.process()?;
         let metrics = run.done();
         Ok((report, metrics))
@@ -355,7 +358,6 @@ impl Engine {
         Ok(())
     }
 }
-
 
 
 //------------ Run -----------------------------------------------------------
@@ -393,7 +395,7 @@ impl<'a, P> Run<'a, P> {
     ) -> Self {
         Run {
             validation, collector, store, processor,
-            metrics: Metrics::new()
+            metrics: Default::default()
         }
     }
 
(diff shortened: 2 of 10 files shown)


src/engine.rs
```diff
@@ -1486,6 +1614,26 @@ impl CaCert {
...
+            Some(parent) => parent,
+            None => return true,
+        };
+
+        match self.rpki_notify() {
+            Some(rpki_notify) => {
+                Some(rpki_notify) != parent.rpki_notify()
+            }
+            None => {
+                self.ca_repository.module() != parent.ca_repository.module()
```
DRiKE, line 1633: Sounds like this is comparing the directories on the (possibly same) rsync server, ergo comparing 'publication points' instead of 'repositories'.
Also, is there a reason not to use .canonical_module() here?

src/engine.rs
```diff
@@ -1598,8 +1863,12 @@ pub trait ProcessCa: Sized + Send + Sync {
     /// it returns `Ok(Some(value))` with a new processor to be used for this
     /// CA. If it wishes to skip this CA, it returns `Ok(None)`. And if it
     /// wishes to abort processing, it returns an error.
+    ///
+    /// The `repository_index` argument indicates the indes of the repository
```
DRiKE, line 1867:
```suggestion
    /// The `repository_index` argument indicates the index of the repository
```

src/engine.rs
```diff
@@ -1582,8 +1840,15 @@ pub trait ProcessRun: Send + Sync {
 
 /// A type that can process the valid data from an RPKI publication point.
 pub trait ProcessCa: Sized + Send + Sync {
+    /// Sets the index of repository in the processing run metrics.
+    fn repository_index(&mut self, repository_index: usize) {
+        let _ = repository_index;
```
DRiKE, line 1845: Not sure what this is supposed to do, is this intentional? Same goes for `fn update_refresh` below.

src/store.rs
```diff
@@ -931,7 +941,10 @@ impl<'a> From<&'a StoredManifest> for sled::IVec {
         // The caRepository URI and manifest bytes are encoded as its bytes
         // preceeded by the length as a u32 in network byte order. the CRL
         // bytes are just the bytes until the end of the buffer.
-        let mut vec = Vec::new();
+        let mut vec = Vec::with_capacity(
+            manifest.ca_repository.as_slice().len() + manifest.manifest.len()
+            + manifest.crl.len() + 17
```
DRiKE, line 946: magic 17 again

src/store.rs
```diff
@@ -1043,7 +1056,10 @@ impl StoredObject {
 
 impl<'a> From<&'a StoredObject> for sled::IVec {
     fn from(object: &'a StoredObject) -> sled::IVec {
-        let mut vec = Vec::new();
+        let mut vec = Vec::with_capacity(
+            object.hash.as_ref().map(|hash| hash.as_slice().len()).unwrap_or(0)
+            + object.content.as_ref().len() + 2
```
DRiKE, line 1061: and magic 2 too

src/origins.rs
```diff
@@ -779,13 +785,24 @@ impl AddressOriginSet {
...
+                    Some(index) => Some(&mut metrics.repositories[index].vrps),
+                    None => None,
+                };
                 for addr in origin.addrs {
-                    tal_metrics.total_valid_vrps += 1;
+                    tal_metrics.valid += 1;
+                    if let Some(vrps) = repo_metrics.as_mut() {
+                        vrps.valid += 1
+                    }
+                    metrics.vrps.valid += 1;
```
DRiKE, line 798: Aren't these counted twice (here, and finally when the metrics are collapsed?)
Goes for the other metrics in this part as well.

src/collector/rrdp.rs
```diff
@@ -1304,7 +1308,7 @@ impl RepositoryState {
 
 impl<'a> From<&'a RepositoryState> for IVec {
     fn from(state: &'a RepositoryState) -> IVec {
-        let mut vec = Vec::new();
+        let mut vec = Vec::with_capacity(state.session.as_bytes().len() + 17);
```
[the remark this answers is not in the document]
partim: Dunno. Isn’t it obvious where it comes from and isn’t it fun to do your own math?

src/engine.rs
partim: This prevents an unused variable warning. Normally you would prefix the variable name with an underscore, but since this is an argument and it shows up in the documentation, I didn’t want to do that.

src/engine.rs
```diff
@@ -991,18 +1046,24 @@ impl<'a, P: ProcessRun> PubPoint<'a, P> {
...
+            self.metrics.invalid_crls += 1;
             return Err(ValidationError)
         }
         if crl.is_stale() {
-            //self.metrics.inc_stale_count();
+            self.metrics.stale_crls += 1;
             match self.run.validation.stale {
                 FilterPolicy::Reject => {
                     warn!("{}: stale CRL.", crl_uri);
+                    self.metrics.invalid_manifests += 1;
```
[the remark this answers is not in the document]
partim: All of these checks are followed by early returns so I think the counting is correct?

src/origins.rs
partim: The engine collapses only the publication metrics while here only VRP-related metrics are collected. I have stuck a note on the collapse function to point this out.

src/engine.rs
partim: The module returned by uri::Rsync::module is the URI up to and including the first path component. Which is what the metrics are relative for.

It is not using canonical module because that is more expensive. Since accidentally claiming a switch doesn’t break anything and differences in case in the host name (which is what canonical module checks for) are rare, this seemed a good trade-off.

src/engine.rs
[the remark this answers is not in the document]
DRiKE: You are right, I think I overlooked those or mixed the stale/invalid counts up.

src/collector/rrdp.rs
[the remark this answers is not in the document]
partim: I’ve decide to leave a bit of mystery in the code.

Seriously, though: the code that explains the length is right below and straightforward, so I feel there is no need to spell it out. For the record: @DRiKE still disagrees. So, impasse.

PR #495, opened 2021-03-29
Clean and update dependencies.

partim:
This removes some unused dependencies and `cargo update`s the whole shebang.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,10 @@ crossbeam-utils = "0.8.1"
 dirs            = "3.0.1"
 fern            = "0.6.0"
 futures         = "0.3.4"
-hyper           = { version = "0.14.2", features = [ "server", "stream" ] }
+
+# Pin hyper to 0.14.4 for compatibility with Rust 1.45.
+hyper           = { version = "=0.14.4", features = [ "server", "stream" ] }
+
 log             = "0.4.8"
 log-reroute     = "0.1.5"
 num_cpus        = "1.12.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -32,7 +34,6 @@ ring            = "0.16.12"
 rpki            = { git = "https://github.com/NLnetLabs/rpki-rs.git", features = [ "repository", "rrdp", "rtr" ] }
 serde           = { version = "1.0.95", features = [ "derive" ] }
 serde_json      = "1.0.57"
-slab            = "0.4.2"
 sled            = "0.34.6"
 tempfile        = "3.1.0"
 tokio           = { version = "1.0", features = [ "io-util", "macros", "rt-multi-thread", "signal", "sync" ] }
```
density215, line 39: Shouldn't tokio be bumped a bit as well? They're currently at 1.4

density215 (approved):

Cargo.toml
partim: "1.0" includes any 1 release. Indeed, Cargo.lock has 1.4.0.

density215: Sure. You might want to signal your intent to use not less than 1.4. OTOH maybe you don't.

partim: Only if there is anything in the newer releases that I totally want. I’ll have a look at the release notes since 1.0.

(I guess allowing the oldest possible version doesn’t matter so much for a binary crate. But then, I am currently cleaning the inner workings up to be also be usable as a library.)

partim: There’s a few improvements to things that our current code works around, so we can probably do an upgrade at a later stage, but nothing that would require anything above Tokio 1.0.0 right now.

PR #506, opened 2021-04-14
Exit the server when a fatal error happens.

partim:
This PR changes the engine the exit its processing loop if an error happens. It also moves the check into the inner loops so the threads exit faster in this case.


DRiKE (approved):
LGTM

PR #507, opened 2021-04-14
Make RRDP fallback random between refresh and rrdp-fallback-time.

### LINKED ISSUE #492: Determine and maintain a random per-RRDP-repository fallback time.
Currently, fallback from RRDP to rsync happens a constant time after the last successful update via RRDP. It has been suggested to instead fall back after a random time to spread out load on the rsync servers.

So, after each RRDP update, not only store the time of update but also a random fallback time with the RRDP repository state and use that to determine whether to fall back to rsync.

partim:
This PR picks and stores a random time between the refresh time and rrdp-fallback-time (which can never be smaller than twice the refresh time) upon each successful update of an RRDP repository. When updating fails, it will fall back to rsync after that time.

Resolves #492.

### DIFF
--- a/src/collector/rrdp.rs
+++ b/src/collector/rrdp.rs
@@ -25,6 +25,7 @@ use chrono::{DateTime, Utc, TimeZone};
 use log::{debug, error, info, warn};
 use reqwest::{Certificate, Proxy, StatusCode};
 use reqwest::blocking::{Client, ClientBuilder, Response};
+use rand::Rng;
 use ring::digest;
 use ring::constant_time::verify_slices_are_equal;
 use rpki::{rrdp, uri};
@@ -71,7 +72,7 @@ pub struct Collector {
     ///
     /// This is the time since the last known update of an RRDP repository
     /// before it is considered non-existant.
-    fallback_time: Duration,
+    fallback_time: FallbackTime,
 }
 
 impl Collector {
@@ -92,7 +93,7 @@ impl Collector {
                 None
             },
             filter_dubious: !config.allow_dubious_hosts,
-            fallback_time: config.rrdp_fallback_time,
+            fallback_time: FallbackTime::from_config(config),
         }))
     }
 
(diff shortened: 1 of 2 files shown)


partim:
Note: This PR changes the format of content in the RRDP database. Because we haven’t released the database code yet, there is no migration. Instead, Routinator will fail and needs to be run once with `--fresh`.

src/collector/rrdp.rs
```diff
@@ -1628,6 +1646,34 @@ impl TryFrom<IVec> for RepositoryObject<Bytes> {
...
+    min: Duration,
+    max: Duration,
+}
+
+impl FallbackTime {
+    /// Creates a new value from the configuration.
+    pub fn from_config(config: &Config) -> Self {
+        FallbackTime {
+            min: config.refresh,
+            max: cmp::min(2 * config.refresh, config.rrdp_fallback_time)
```
DRiKE, line 1663:
```suggestion
            max: cmp::max(2 * config.refresh, config.rrdp_fallback_time)
```

DRiKE: Using `min` would allow a `config.rrdp_fallback_time` lower than `2 * config.refresh`, which is not what you want according to the documentation

src/collector/rrdp.rs
```diff
@@ -1628,6 +1646,34 @@ impl TryFrom<IVec> for RepositoryObject<Bytes> {
...
+        }
+    }
+
+    /// Picks a best-before date for a repository updated around now.
+    pub fn best_before(self) -> DateTime<Utc> {
+        // Saturating conversion between std’s and chrono’s Duration types.
+        Utc::now() + chrono::Duration::from_std(
+            rand::thread_rng().gen_range(self.min..self.max)
+        ).unwrap_or_else(|_| chrono::Duration::milliseconds(i64::MAX))
```
DRiKE, line 1672: Does this `_or_else` mean that in case anything goes wrong, the best_before date will be very far in the future and the fallback to rsync will realistically never happen? Why not default back to half an hour or an hour worth of milliseconds, in that error case?

DRiKE (changes requested):

src/collector/rrdp.rs
partim: The conversation can only go wrong if the std duration is too big. So this just cuts it back to whatever the biggest value is. Not perfect but also very theoretical.

DRiKE (approved):

PR #514, opened 2021-04-19
Allow input for the validate command from and output to a file.

### LINKED ISSUE #430: Allow file input and output for prefix validation
We were trying to increase the interval between consecutive refreshes (--refresh) to a large value for testing purposes. It seems there is an upper bound to the length of intervals: the interval will always be around 1.5 hours even if the value of the flag is way larger than 1.5 hours (we used 86400 seconds in our experiments). I monitored the interval length by checking the timestamps in /status.

partim:
This PR adds a new `--input` option to the `validate` command to read a list of route announcements from a file or standard input. It reads either JSON or a simple plain text format. It also adds `--output` to redirect output to a file.

Resolves #430.


doc/routinator.1
```diff
@@ -550,24 +550,71 @@ are provided. It can also be combined with one or more ASN filters. Then all
 matching VRPs are included. That is, filters combine as "or" not "and."
 
 .SS validate
-This command can be used to perform RPKI route origin validation for a route
-announcement. Routinator will determine whether the provided announcement is
-RPKI valid, invalid, or not found.
+This command can be used to perform RPKI route origin validation for one or
+more route announcement. Routinator will determine whether the provided announcements are RPKI valid, invalid, or not found.
```
AlexanderBand, line 554: s/announcement/announcements

DRiKE (approved):

PR #523, opened 2021-04-22
Fix delta generation. Increase serial number.

partim:
Let’s not talk about this one …

### DIFF
--- a/src/payload.rs
+++ b/src/payload.rs
@@ -1606,8 +1606,8 @@ mod test {
                 &HashMap::from_iter(
                     vec![(2, ()), (4, ()), (5, ())].into_iter()
                 )
-            ),
-            vec!(1, 3),
+            ).into_iter().collect::<HashSet<_>>(),
+            HashSet::from_iter(vec![1, 3].into_iter()),
         );
     }
 }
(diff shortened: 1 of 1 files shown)


src/payload.rs
```diff
@@ -1519,7 +1529,7 @@ impl RoaInfo {
 fn key_difference<K: Copy + Hash + Eq, V>(
```
DRiKE, line 1529: Nit/check: this should (also looking at the test below) return keys from `current` not occurring in `next`, but _not_ the other way around? Perhaps the comment describing the function should be more detailed, especially because the function name is generic (`_difference`)

DRiKE (approved):

src/payload.rs
partim: I have renamed the function to better reflect what it actually does. I also added a test for delta construction.

src/payload.rs
```diff
@@ -1595,11 +1595,58 @@ mod test {
...
+        next.origins.insert(o4, info.clone());
+        let delta = PayloadDelta::construct(
+            &current, &next, 12.into()
+        ).unwrap();
+
+        assert_eq!(delta.serial, Serial::from(13));
+        let mut add: HashSet<_> = delta.announced_origins.into_iter().collect();
+        let mut sub: HashSet<_> = delta.withdrawn_origins.into_iter().collect();
+
+        eprintln!("{:?}", add);
```
DRiKE, line 1630: Seems to be a (harmless, likely) leftover debug statement?

DRiKE (approved):

PR #527, opened 2021-04-28
Do not keep a collector in no-update mode.

partim:
This PR changes how we deal with the `--no-update` option available in some commands: We simply don’t use the collector at all. It changes the collector field in the engine to an option and adjusts various other types and methods to be able to cope with that.


src/engine.rs
```diff
@@ -348,13 +361,19 @@ impl Engine {
 
     /// Cleans the collector and store owned by the engine.
     pub fn cleanup(&self) -> Result<(), Failed> {
-        self.store.cleanup(self.collector.cleanup())
+        self.store.cleanup(
+            self.collector.as_ref().map(|collector| {
```
DRiKE, line 365: Any reason this can't be written as `as_ref().map(Collector::cleanup)`, to be shorter but moreover consistent with line 343?

DRiKE (approved):

src/engine.rs
partim: Apparently, no reason. I am surprised Clippy didn’t complain.

PR #529, opened 2021-04-29
Log falling back to rsync only once per repository.

partim:
This PR fixes an annoyance where the message that a fallback to rsync had to happen is logged every single time a publication point from the affected repository is accessed.

### DIFF
--- a/src/collector/base.rs
+++ b/src/collector/base.rs
@@ -6,7 +6,7 @@ use std::{fs, io};
 use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 use bytes::Bytes;
-use log::{error, warn};
+use log::error;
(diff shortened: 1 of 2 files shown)


src/collector/base.rs
```diff
@@ -215,10 +215,6 @@ impl<'a> Run<'a> {
                 if let Some(repository) = rrdp.load_repository(rrdp_uri)? {
                     return Ok(Some(Repository::rrdp(repository)))
                 }
-                warn!(
```
DRiKE, line 218: The comment above this function needs an update, as with the removal of this `warn!` there is no diagnostic information logged at all anymore (in this place).

DRiKE (approved):

PR #531, opened 2021-04-29
Limit maximum size of downloaded objects.

### LINKED ISSUE #462: Allow limiting the size of objects downloaded via RRDP.
Also look into whether this can be done to rsync via some sort of command line option.

partim:
This PR introduces the new config option `max-object-size` that limits the size of individual RPKI objects. It is enabled by default and set to 20 Mbytes. The largest object currently in the RPKI – RIPE’s manifest – is a bit over 1 Mbyte, so this should give a bit of buffer.

The limit is implemented in the RRDP client which simply stops processing and errors out when any published object exceeds the size limit. For rsync, we pass the `--max-size` option if no other options are provided.

Resolves #462


ties:
I would also consider the AS0 ROA. That used to be quite large (megabytes) as well.

partim:
I checked APNIC’s AS0 ROA. It is about a megabyte, so comparable with RIPE’s manifest.

src/collector/rrdp.rs
```diff
@@ -1956,6 +1983,58 @@ impl<R: io::Read> io::Read for HashRead<R> {
...
+
+impl<R: io::Read> io::Read for MaxSizeRead<R> {
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize, io::Error> {
+        let res = self.reader.read(buf)?;
+        if let Some(left) = self.left {
+            let res64 = match u64::try_from(res) {
+                Ok(res) => res,
+                Err(_) => {
+                    // If the usize doesn’t fit into a u64, things are
+                    // definitiely way too big.
```
DRiKE, line 2013:
```suggestion
                    // definitely way too big.
```

src/config.rs
```diff
@@ -392,6 +398,12 @@ impl Config {
             .help("Keep RRDP responses in the given directory")
             .takes_value(true)
         )
+        .arg(Arg::with_name("max-object-size")
+            .long("max-object-size")
+            .value_name("BYTES")
+            .help("Maximum size of downloaded objects (0 .. no limit)")
```
DRiKE, line 404: This help message seems strange, like it allows to pass anything between 0 and unlimited, while 0 actually should be used for unlimited. Something like `(0 means no limit)` perhaps?

doc/routinator.1
```diff
@@ -313,6 +313,12 @@ This option provides the URI of a proxy to use for all HTTP connections made
 by the RRDP client. It can be either an HTTP or a SOCKS URI. The option can
 be given multiple times in which case proxies are tried in the given order.
 
+.TP
+.BI --max-object-size= BYTES
+Limits the size of individual objects received via either rsync or RRDP to
+the given number of bytes. The default value if this option is not present
+is 20 MBytes. Use a value of 0 to disable the limit.
```
DRiKE, line 320: I could see users getting confused by the '20', thinking this option takes Megabytes instead of bytes and perhaps misconfigure their setup into a breaking way.
```suggestion
is 20000000 (i.e. 20 MBytes). Use a value of 0 to disable the limit.
```

doc/routinator.1
```diff
@@ -1041,6 +1047,12 @@ A list of string each providing the URI for a proxy for outgoing RRDP
 connections. The proxies are tried in order for each request. HTTP and SOCKS5
 proxies are supported.
 
+.TP
+.B max-object-size
+An integer value that provides a limit for the size of individual objects
+received via either rsync or RRDP to the given number of bytes. The default value if this option is not present is 20 MBytes. A value of 0 disables
```
DRiKE, line 1053:
```suggestion
received via either rsync or RRDP to the given number of bytes. The default value if this option is not present is 20000000 (i.e. 20 MBytes). A value of 0 disables
```

DRiKE (approved):

PR #537, opened 2021-05-06
Provide delta output in JSON under /json-delta.

### LINKED ISSUE #467: Offer RTR-like access to VRPs via HTTP/JSON.

partim:
This PR provides a RTR-like delta output in JSON as the HTTP /json-delta endpoint.

Resolves #467


src/http/delta.rs
```diff
@@ -0,0 +1,327 @@
...
+            origin.max_length()
+        ).unwrap()
+    }
+
+    /// Appends the footer to the vec.
+    fn append_footer(vec: &mut Vec<u8>) {
+        vec.extend_from_slice(b"\n  ]\n}\n");
+    }
+}
+
```
density215, line 201: Any particular reason for not using Serde here?

src/http/errors.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+pub fn not_found() -> Response<Body> {
+    Response::builder()
+    .status(StatusCode::NOT_FOUND)
+    .header("Content-Type", "text/plain")
+    .body("Not Found".into())
+    .unwrap()
+}
+
```
density215, line 37: Don't know if this applicable here, but as an API consumer I prefer to have error responses also be of type 'application/json' and with the error message as JSON (well, except for 500).

OTOH probably everybody is dropping the whole thing after reading the HTTP Status code and never bother to parse the error message, also since it only re-iterates the error code.

src/http/delta.rs
```diff
@@ -0,0 +1,327 @@
...
+) -> Response<Body> {
+    Response::builder()
+    .header("Content-Type", "application/json")
+    .body(Body::wrap_stream(stream::iter(
+        SnapshotStream::new(session, to_serial, snapshot)
+        .map(Result::<_, Infallible>::Ok)
+    )))
+    .unwrap()
+}
+
```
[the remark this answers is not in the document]
partim: I have an issue somewhere for replacing those. It was initially part of 0.9, but I have punted it to 0.9.1 because I want to look at all the unwraps everywhere and replace those if at all possible.

src/http/delta.rs
partim: Yes: This implementation allows to stream the result in blocks to save memory. I don’t think this can be done with serde.

src/http/errors.rs
partim: Hm, this is a bit tricky. I guess we could do all error messages under the `/api` tree as JSON? I kind of want to keep the ones where you might actually go with your browser as plain text.

density215: Under `/api` only makes sense, I guess

partim: After toying with possible solutions, I don’t think the effort is worth it. I think I’ll just leave it as it is.

PR #539, opened 2021-05-07
Clean up and extend Prometheus metrics.

### LINKED ISSUE #494: Clean up Prometheus metrics.
Add newly introduced metrics to the Prometheus endpoint. Look into making maintenance of that endpoint less tedious and error prone.

partim:
This PR adds the new metrics to the Prometheus metrics output and uses the opportunity to sanitize generation of these metrics a bit.

Resolves #494

### DIFF
--- a/src/http/status.rs
+++ b/src/http/status.rs
@@ -38,6 +38,8 @@ async fn handle_metrics(
     server_metrics: &HttpServerMetrics,
     rtr_metrics: &SharedRtrServerMetrics,
 ) -> Response<Body> {
+    use PrometheusType::*;
+
     let (metrics, serial, start, done, duration) = {
         let history = history.read();
         (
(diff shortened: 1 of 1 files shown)


src/http/status.rs
```diff
@@ -51,328 +51,215 @@ async fn handle_metrics(
...
-                + f64::from(duration.subsec_millis()) / 1000.
-            ).unwrap();
+        );
+        if let Ok(duration) = rrdp.duration {
+            res.multi(
+                "rrdp_duration", "uri", &rrdp.notify_uri,
+                format_args!(
+                    "{:.3}",
+                    duration.as_secs() as f64
+                    + f64::from(duration.subsec_millis()) / 1000.
```
DRiKE, line 179: inconsistent way of formatting, see comment below

src/http/status.rs
```diff
@@ -51,328 +51,215 @@ async fn handle_metrics(
...
+                Err(_) => -1
+            }
+        );
+        if let Ok(duration) = rsync.duration {
+            res.multi(
+                "rsync_duration", "uri", &rsync.module, 
+                format_args!(
+                    "{:.3}",
+                    duration.as_secs() as f64
+                    + f64::from(duration.subsec_millis()) / 1000.
```
DRiKE, line 214: inconsistent way of formatting, see comment below

src/http/status.rs
```diff
@@ -406,113 +290,350 @@ async fn handle_metrics(
...
-                        duration.num_seconds(),
-                        duration.num_milliseconds() % 1000,
-                    ).unwrap();
+                    res.multi(
+                        "rtr_client_last_update_seconds",
+                        "addr", addr,
+                        format_args!(
+                            "{}.{:03}",
+                            duration.num_seconds(),
+                            duration.num_milliseconds() % 1000,
```
DRiKE, line 302: inconsistent with the other duration calculations (`rrdp_duration`, `rsync_duration`, lines 179 and 214 respectively).
IMO the code here is just a bit cleaner / clearer than in the other parts.

src/http/status.rs
```diff
@@ -406,113 +290,350 @@ async fn handle_metrics(
...
-        # TYPE routinator_http_current_connections gauge"
-    ).unwrap();
-    writeln!(res,
-        "routinator_http_current_connections {}",
+    res.header(
+        "http_current_connections", "gauge",
+        "number of currently open HTTP connections"
+    );
+    res.single(
+        "http_current_onnections",
```
DRiKE, line 355:
```suggestion
        "http_current_connections",
```

src/http/status.rs
```diff
@@ -406,113 +290,350 @@ async fn handle_metrics(
...
+            "stray_crls", "gauge",
+            format_args!("stray CRLs {}", group)
+        );
+        self.header(
+            "valid_ca_certs", "gauge",
+            format_args!("valid CA certificates {}", group)
+        );
+        self.header(
+            "valid_ee_certs", "gauge",
+            format_args!("valid router certificates {}", group)
```
DRiKE, line 435: router certificates? or ee certificates?

src/http/status.rs
```diff
@@ -406,113 +290,350 @@ async fn handle_metrics(
...
+        );
+        self.header(
+            "invalid_certs", "gauge",
+            format_args!("invalid certificate files {}", group)
```
DRiKE, line 439: consistency wrt the `valid_*_certs` headers :
```suggestion
            format_args!("invalid certificates {}", group)
```

src/http/status.rs
```diff
@@ -406,113 +290,350 @@ async fn handle_metrics(
...
+
+    fn vrp_headers(&mut self, group: &str) {
+        self.header(
+            "vrps_total", "gauge",
+            format_args!("total number of encountered valid VRPs {}", group)
+        );
+        self.header(
+            "vrps_unsafe", "gauge",
+            format_args!(
+                "VRPs overlapping with rejected publication points  {}",
```
DRiKE, line 549:
```suggestion
                "VRPs overlapping with rejected publication points {}",
```

src/http/status.rs
```diff
@@ -51,328 +51,215 @@ async fn handle_metrics(
...
-        ).unwrap();
+    res.header(
+        "stale_objects", "gauge",
+        "total number of stale manifests and CRLs"
+    );
+    res.single("stale_objects", metrics.publication.stale_objects());
+
+    // Per-repository metrics.
+    res.set_prefix("routinator_repo");
+    res.set_prefix("routinator");
```
DRiKE, line 82: and this is the default value set by `PrometheusMetrics::new`, so this one can go as well

DRiKE (changes requested):
There are some typos in what would end up in the final prometheus output, which obviously need to be fixed. Perhaps a prometheus linting tool, if such thing exist, would be a nice extra check.

src/http/status.rs
partim: All EE certificates are router certificates. I felt that term to be more clear than some inside abbreviation.

src/http/status.rs
partim: This was intentional to make clear that this does only count invalid stand-alone certificates rather than those certificates included in other objects (like ROAs).

src/http/status.rs
partim: This was intended to be _after_ the pre-repository metrics to reset the prefix to its default. This way `publication_values` and `vrp_values` methods can be used both for the TA and repository metrics.

AlexanderBand (approved):

DRiKE (approved):

src/http/status.rs
ties: This tripped me up in the beta web ui: I think of embedded EE certificates as EE certificates as well.

partim: Good point. We’re looking into changing it into ‘router certificates’ there, too.

PR #566, opened 2021-05-26
Build Routinator RPM packages

ximon18:
This is a draft PR at present because I want to add to this PR a GitHub Actions workflow that uses Docker and LXC like the Debian packaging does to actually build and test the RPMs.


ximon18:
I will mark this PR as ready for review to see if that causes GH to run the new RPM packaging workflow. If not I might have to push a copy of the workflow file to `main` as well.

ximon18:
Progress: the https://github.com/NLnetLabs/routinator/pull/566/checks?check_run_id=2679469756 build passed.

The CentOS 7 package needs to be built with the minimal systemd unit, as we do for older Debian versions. It isn't currently failing but the logs on the client on which Routinator is installed contain a lot of warnings about unrecognized configuration keys in the service unit file.

ximon18:
> The CentOS 7 package needs to be built with the minimal systemd unit, as we do for older Debian versions. It isn't currently failing but the logs on the client on which Routinator is installed contain a lot of warnings about unrecognized configuration keys in the service unit file.

This is now handled by the GitHub Actions workflow.

PR #568, opened 2021-05-27
Update minimal Rust version to 1.47. 

partim:
This version is required to allow removing the pin on hyper. As Alpine 3.13 now has updated to a newer Rust version, we can update.

This also updates the Dockerfile to use Alpine 3.13.5 for both stages. Previously, the second stage was still using 3.11 (Alpine, not Windows) for some reason.


ties:
Iirc the krill dockerfile was using an arg to make sure that the same container was used. You may want to copy that trick here to prevent them diverging again.

ximon18 (approved):

PR #570, opened 2021-05-31
Add RHEL/CentOS Package support

AlexanderBand:


README.md
```diff
@@ -67,6 +67,43 @@ configuration file located in `/etc/routinator/routinator.conf`. You can check
 the status of Routinator with `sudo systemctl status routinator` and view the
 logs with `sudo journalctl --unit=routinator`.
 
+## Quick Start with RHEL/CentOS Packages
+
+Assuming you have a machine running a RHEL/CentOS 7 or 8 distribution, you can 
+install Routinator, you can install Routinator from our [software package
```
ximon18, line 73: "you can install Routinator, you can install Routinator" - no need to say it twice ;-)

README.md
```diff
@@ -67,6 +67,43 @@ configuration file located in `/etc/routinator/routinator.conf`. You can check
...
+baseurl=https://packages.nlnetlabs.nl/linux/centos/$releasever/main/$basearch
+enabled=1
+```
+Then run the following command to add the public key
+
+```bash
+sudo rpm --import https://packages.nlnetlabs.nl/aptkey.asc
+```
+
+You can then install, initialise, enable and start Routinator by running these
```
ximon18, line 90: From this point on it's the same as for DEBs, is it worth stating this only once for both?

AlexanderBand: I'm gonna leave it for now and think about what we want in the README now that we have docs like they are.

partim (approved):

ximon18 (approved):

PR #606, opened 2021-07-19
Fix dump logic.

### LINKED ISSUE #604: Dumping stored data has incorrect structure
Currently a `stored` directory is created with all RRDP publication points as a separate subfolder directly under it. 

- The parent folder should be called `store` instead
- All RRDP respositories should be in an `rrdp` subfolder (`rsync` does have its own subfolder).

Screenshot 2021-07-19 at 15 57 15

partim:
That apparently got broken during the Second Big Rewrite. This PR fixes triggering the collector dump in the dump command and corrects various wrong paths in the RRDP collector’s dump.

Fixes #604

### DIFF
--- a/src/collector/rrdp.rs
+++ b/src/collector/rrdp.rs
@@ -142,8 +142,9 @@ impl Collector {
     }
 
     /// Dumps the content of the RRDP collector.
-    #[allow(clippy::mutable_key_type)]
     pub fn dump(&self, dir: &Path) -> Result<(), Failed> {
+        let dir = dir.join("rrdp");
+        debug!("Dumping RRDP collector content to {}", dir.display());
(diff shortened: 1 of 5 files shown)


DRiKE (approved):
The directory structure (with the 'rsync' subdir for rrdp things) keeps confusing me but I think what's going on here is correct. If possible, it might be nice to get rid of the manual path joining (i.e. appending 'rsync', 'rrdp') in multiple places.

partim:
> If possible, it might be nice to get rid of the manual path joining (i.e. appending 'rsync', 'rrdp') in multiple places.

I hear ya, but I think in this case it sort of is justified and would just result in adding about ten lines of code without too much benefit.

PR #608, opened 2021-07-19
Check for short maxlen in SLURM prefix assertions.

### LINKED ISSUE #584: Ensure the max-length isn’t smaller than the prefix lenght in SLURM files.
Currently, this is not checked _anywhere_ and may result in corrupt VRPs rejected by routers.

partim:
Fixes #584.

### DIFF
--- a/src/slurm.rs
+++ b/src/slurm.rs
@@ -290,7 +290,7 @@ struct PrefixAssertion {
     comment: Option<String>,
 }
 
-// We neeed to enforce that max_prefix_len is greater or equal to
+// We need to enforce that max_prefix_len is greater than or equal to
 // prefix.len(), so we need to roll our own implementation.
 impl<'de> Deserialize<'de> for PrefixAssertion {
     fn deserialize<D: Deserializer<'de>>(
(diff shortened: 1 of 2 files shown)


src/slurm.rs
```diff
@@ -279,19 +279,121 @@ struct BgpsecFilter {
...
 
     asn: u32,
 
-    #[serde(rename = "maxPrefixLength")]
     max_prefix_len: Option<u8>,
 
     comment: Option<String>,
 }
 
+// We neeed to enforce that max_prefix_len is greater or equal to
```
DRiKE, line 293:
```suggestion
// We need to enforce that max_prefix_len is greater than or equal to
```

test/slurm/bad_maxlen.json
```diff
@@ -0,0 +1,59 @@
...
+      }
+    ]
+  },
+  "locallyAddedAssertions": {
+    "prefixAssertions": [
+      {
+        "asn": 64496,
+        "prefix": "198.51.100.0/24",
+        "maxPrefixLength": 20,
+        "comment": "My other important route"
```
DRiKE, line 41: Nit, but perhaps be explicit that this is the rule that should trigger the `is_err` in the new `[test]`
```suggestion
        "comment": "My other important route with invalid short maxlen"
```

DRiKE (approved):

src/slurm.rs
partim: No, but we reaaaaalllly do …

PR #612, opened 2021-07-20
Increase the default RRDP timeout to 300 seconds.

### LINKED ISSUE #595: Increase default RRDP timeout.
The default RRDP timeout is currently 30 seconds whereas the default rsync timeout is 300 seconds. Given that we keep seeing timeouts on RRDP for the APNIC repository in certain parts of Europe, increasing the RRDP timeout to at least 300 seconds seems apt.

partim:
This also cleans up the awkward `Option<Option<Duration>>` in the config for this value.

Fixes #595.


DRiKE (approved):
The changes seem fine, however the difference between what `rrdp_timeout` and `rrdp_connect_timeout` exactly affect is not quite clear from the code/comments itself (though the man page explains the difference).

PR #614, opened 2021-07-20
Include the rrdp-fallback-time command line option.

partim:
‘Someone’ forgot to add the option to the command line parser. This PR fixes that.


src/config.rs
```diff
@@ -364,6 +364,12 @@ impl Config {
             .long("disable-rrdp")
             .help("Disable RRDP and only use rsync")
         )
+        .arg(Arg::with_name("rrdp-fallback-time")
+            .long("rrdp-fallback-time")
+            .takes_value(true)
+            .value_name("SECONDS")
+            .help("Maximum age of an RRDP repository before fallback to rsync")
```
DRiKE, line 371: 'Age' sounds very strange to me, though it's challenging to come up with something that is as brief. Suggestion that is a bit more consistent with the man page:
```suggestion
            .help("Maximum time since last successful update before fallback to rsync")
```

DRiKE (approved):

PR #617, opened 2021-07-21
Log the VRP summary at the end of a server validation run.

### LINKED ISSUE #616: Log the summary at the end of a run
Basically the output of `-f summary`.

partim:
Fixes #616.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -5,7 +5,7 @@ use std::str::FromStr;
 use std::sync::Arc;
 use chrono::Utc;
 use chrono::format::{Item, Numeric, Pad};
-use log::error;
+use log::{error, info};
 use rpki::repository::resources::AsId;
 use crate::payload;
 use crate::error::Failed;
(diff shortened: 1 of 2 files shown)


src/output.rs
```diff
@@ -816,6 +817,45 @@ impl<W: io::Write> Formatter<W> for Summary {
...
+            metrics.vrps.valid, metrics.vrps.marked_unsafe,
+            metrics.vrps.contributed,
+        ))
+    }
+
+    pub fn log(metrics: &Metrics) {
+        Self::produce_header(metrics, |args| {
+            info!("{}", args);
+            Ok(())
+        }).unwrap()
```
DRiKE, line 846: If for some very exotic reason we can not write to stdout (or, to wherever stdout is redirected), this `unwrap` can cause trouble I guess?

src/output.rs
```diff
@@ -792,11 +792,12 @@ impl<W: io::Write> Formatter<W> for Rpsl {
...
 
-struct Summary;
+/// Output only a summary.
+pub struct Summary;
 
-impl<W: io::Write> Formatter<W> for Summary {
-    fn header(
-        &self, _snapshot: &PayloadSnapshot, metrics: &Metrics, target: &mut W
+impl Summary {
+    pub fn write(
```
DRiKE, line 799: is `write` still used somewhere? Looks like `log` + `produce_header` replaced it entirely.

DRiKE (approved):
No blocking problems, but it seems there is some leftover code (`pub fn write`) or a case of code repetition. Or I am missing something.

src/output.rs
partim: The only way that `produce_header` can return an error is if the closure returns an error, which, in this case it never does, so this is fine.

src/output.rs
partim: Uhm, indeed. Forgot to delete it. Fixed.

PR #678, opened 2022-01-17
Switch to using payload types from rpki 0.14.

partim:
This PR contains the changes necessary to upgrade to _rpki_ 0.14.

Most importantly, we are now using `rpki::rtr::payload::Payload` everywhere. Since that type also includes BGPsec router keys, it includes a number of preparations for router key support. However, at this point we still only support route origins.

The PR also switches SLURM support to the types provided by the _rpki_ crate. There should be no consequences from that.

Before it can be merged, I would like to add a bunch of test cases to at least the `payload` to make sure the changes didn’t break anything.


partim:
Comparing the output between this PR and 0.10.2 shows the exact same numbers, so I have decided to punt the tests until the adjustments for BGPsec keys. That’s way more interesting, anyway.

Cargo.toml
```diff
@@ -29,7 +29,8 @@ num_cpus        = "1.12.0"
 rand            = "0.8.1"
 reqwest         = { version = "0.11.0", default-features = false, features = ["blocking", "rustls-tls" ] }
 ring            = "0.16.12"
-rpki            = { version = "0.12.2", features = [ "repository", "rrdp", "rtr", "serde" ] }
+routecore       = "0.1.1"
+rpki            = { git = "https://github.com/NLnetLabs/rpki-rs.git", branch = "main", features = [ "repository", "rrdp", "rtr", "serde", "slurm" ] }
```
density215, line 33: Shouldn't this be  = "0.14.1"?

partim: I had to add a few things to rpki to make it work with Routinator, so we need the main branch until there is 0.14.2.

density215 (approved):

PR #681, opened 2022-01-18
Reject premature manifests.

### LINKED ISSUE #589: Reject premature manifests.
Draft 6489bis requires that a manifest before its thisUpdate time be rejected as a well. Do this, and add the appropriate metrics.

partim:
Manifests are considered premature if their thisUpdate timestamp is
before the current time. Internet draft draft-ietf-sidrops-6486bis added
a requirement to reject such manifests. This PR adds this requirement.

A new state "premature" has been added to the manifest Prometheus
metrics. Similarly, a new prematureManifests counter has been added to
the JSON metrics.

Fixes #589.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -822,6 +822,13 @@ impl<'a, P: ProcessRun> PubPoint<'a, P> {
                 return Ok(None)
             }
         };
+
+        if content.this_update() < Time::now() {
+            self.metrics.premature_manifests += 1;
+            warn!("{}: premature manifest", self.cert.rpki_manifest());
+            return Ok(None)
+        }
+
         if content.is_stale() {
             self.metrics.stale_manifests += 1;
             match self.run.validation.stale {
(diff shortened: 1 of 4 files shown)


density215 (approved):

src/engine.rs
```diff
@@ -822,6 +822,13 @@ impl<'a, P: ProcessRun> PubPoint<'a, P> {
                 return Ok(None)
             }
         };
+
+        if content.this_update() < Time::now() {
+            self.metrics.premature_manifests += 1;
+            warn!("{}: premature manifest", self.cert.rpki_manifest());
+            return Ok(None)
```
timbru, line 829: Looks consistent with other checks.

That said, I lack some context I suppose but it's not clear to me why this function returns `Result<Option<ValidPointManifest>, Failed>`. As far as I can tell no Err(Failed) is ever returned. So, this feels more like an infallible function to determine whether a given manifest is eligible for use.

timbru (approved):

src/engine.rs
partim: Actually, it calls `validate_collected_crl` which can error.

PR #683, opened 2022-01-18
Count collected valid CRLs.

### LINKED ISSUE #645: Valid CRL metrics seemingly broken
The "Valid CRL" metric seems to display incorrect numbers intermittently.
```
routinator_repository_objects_total{type="crl", state="valid"}
```
This behavior did not occur in 0.9.

Screenshot 2021-09-06 at 16 04 01

partim:
Fixes #645.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -967,6 +967,7 @@ impl<'a, P: ProcessRun> PubPoint<'a, P> {
         }
 
         // Phew: All good.
+        self.metrics.valid_crls += 1;
         Ok(Some((crl_uri, crl, crl_bytes)))
     }
 


density215 (approved):

ties:
Do you have an estimate of when this will be incorporated in a release? This would allow me to switch to 0.10 (which is right now not possible due to spurious alerts on crl count) and prevent other spurious alerts (from a rsync only instance that rejected the a top level manifest after an rsync timeout that just woke me up).

partim:
Without promising anything, it’s looking like the first RC of the next release might be feasible as early as mid next week.

Issue #579, opened 2021-06-15
IPv4 and IPv6 ROA count


csdali13:
Hi,

Could you please let me know is there a way we can get the IPv4 and IPv6 ROA count for the purpose of monitoring in Grafana? I could only find per tal number of ROAs (routinator_valid_roas) in the list of metrics available.

Thanks,
Dali

partim:
We currently do not distinguish between IPv4 and IPv6 prefixes in the metrics. This could, of course, be added in one of the next versions, if desired.

PR #693, opened 2022-01-24
Add support for router keys.

partim:
This PR adds support for BGPsec router keys.

Since there is hardly any practical experience with router keys appearing in RPKI payload sets, the feature will be disabled by default and needs to be switched on via the new `enable-bgpsec` command line or config file option.

Output of router keys has been added to the jsonext format. The other formats will remain unchanged for now. In a follow-up PR we will add JSON output using the SLURM specification (#696).

### DIFF
--- a/doc/manual/source/manual-page.rst
+++ b/doc/manual/source/manual-page.rst
@@ -462,11 +462,15 @@ These can be requested by providing different commands on the command line.
                   format.
 
            jsonext
-                  The list is placed into a JSON object with two members:
+                  The list is placed into a JSON object with three members:
                   *roas* contains the validated route origin
-                  authorizations and *routerKeys* contains the validated 
-                  BGPsec router keys. This second member is also present when
-                  BGPsec has not been enabled, it will simply be always empty.
+                  authorizations, *routerKeys* contains the validated 
+                  BGPsec router keys, and *metadata* contains some infromation
+                  about the validation run itself.
+
+                  All three members are always present, even if BGPsec has
+                  not been enabled. In this case, *routerKeys* will simply
+                  be empty.
 
                   The *roas* member contains an array of objects with four
                   elements each: The autonomous system number of the network
--- a/doc/manual/source/manual-page.rst
+++ b/doc/manual/source/manual-page.rst
@@ -281,6 +281,11 @@ The available options are:
       The maximum number of CAs a given CA may be away from a trust anchor
       certificate before it is rejected. The default value is 32.
 
+.. option:: --enable-bgpsec
+
+      If this option is present, BGPsec router keys will be processed
+      during validation and included in the produced data set.
+
 .. option:: --dirty
 
       If this option is present, unused files and directories will not be
@@ -1061,6 +1066,11 @@ All values can be overridden via the command line options.
             An integer value that specifies the maximum number of CAs a given
             CA may be away from a trust anchor certificate before it is
             rejected. If the option is missing, a default of 32 will be used.
+
+      enable-dnssec
+            A boolean value specifying whether BGPsec router keys should be
+            included in the published dataset. If false or missing, no router
+            keys will be included.
 
       dirty
             A boolean value which, if true, specifies that unused files and
--- a/doc/manual/source/manual-page.rst
+++ b/doc/manual/source/manual-page.rst
@@ -462,24 +462,41 @@ These can be requested by providing different commands on the command line.
                   format.
 
            jsonext
-                  The list is placed into a JSON object with a single element
-                  *roas* which contains an array of objects with four
+                  The list is placed into a JSON object with two members:
+                  *roas* contains the validated route origin
+                  authorizations and *routerKeys* contains the validated 
+                  BGPsec router keys. This second member is also present when
+                  BGPsec has not been enabled, it will simply be always empty.
+
+                  The *roas* member contains an array of objects with four
                   elements each: The autonomous system number of the network
                   authorized to originate a prefix in *asn*, the prefix in
                   slash notation in *prefix*, the maximum prefix length of
-                  the announced route in *maxLength*.
+                  the announced route in *maxLength*, and extended
+                  information about the source of the authorization in
+                  *source*. 
 
-                  Extensive information about the source of the object is
-                  given in the array *source*. Each item in that array is an
-                  object providing details of a source of the VRP. The object
-                  will have a *type* of *roa* if it was derived from a valid
-                  ROA object or *exception* if it was an assertion in a local
-                  exception file.
+                  The *routerKeys* member contains an array of objects with
+                  four elements each: The autonomous system using the router
+                  key is given in *asn*, the key identifier as a string of
+                  hexadecimal digits in *SKI*, the actual public key as a
+                  Base 64 encoded string in *routerPublicKey*, and extended
+                  information about the source of the key is contained in
+                  *source*.
 
-                  For ROAs, *uri* provides the rsync URI of the ROA,
-                  *validity* provides the validity of the ROA itself, and
-                  *chainValidity* the validity considering the validity of
-                  the certificates along the validation chain.
+                  This source information the same for route origins and
+                  router keys. It consists of an array. Each item in that
+                  array is an object providing details of a source.
+                  The object will have a *type* of *roa* if it was derived
+                  from a valid ROA object, *cer* if it was derived from
+                  a published router certificate, or *exception* if it was an
+                  assertion in a local exception file.
+
+                  For RPKI objects, *uri* provides the rsync URI of the ROA
+                  or router certificate, *validity* provides the validity of
+                  the ROA itself, and *chainValidity* the validity
+                  considering the validity of the certificates along the
+                  validation chain.
 
                   For  assertions from local exceptions, *path* will provide
                   the path of the local exceptions file and, optionally,
--- a/doc/manual/source/manual-page.rst
+++ b/doc/manual/source/manual-page.rst
@@ -1067,7 +1067,7 @@ All values can be overridden via the command line options.
             CA may be away from a trust anchor certificate before it is
             rejected. If the option is missing, a default of 32 will be used.
 
-      enable-dnssec
+      enable-bgpsec
             A boolean value specifying whether BGPsec router keys should be
             included in the published dataset. If false or missing, no router
             keys will be included.
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -45,7 +45,6 @@ use crate::error::Failed;
 use crate::metrics::{
     Metrics, PublicationMetrics, RepositoryMetrics, TalMetrics
 };
-use crate::payload::ValidationReport;
 use crate::store::{Store, StoredManifest, StoredObject, StoredPoint};
 use crate::utils::str::str_from_ascii;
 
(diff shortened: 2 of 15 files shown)


partim:
Regarding metrics: These are already there and are included in the JSON status output. However, I am currently not quite sure how to best model the Prometheus metrics. Any opinion would be appreciated.

Relevant JSON status looks like this:
```
{
   "tals": {
      "ripe": {
         "payload": {
            "routeOriginsIPv4": {
               "total": 127786,
               "unsafe": 17,
               "locallyFiltered": 0,
               "duplicate": 2,
               "final": 127784
            },
            "routeOriginsIPv6": {
               "total": 20948,
               "unsafe": 0,
               "locallyFiltered": 0,
               "duplicate": 0,
               "final": 20948
            },
            "routerKeys": {
               "total": 2,
               "locallyFiltered": 0,
               "duplicate": 0,
               "final": 2
            }
         }
    }
}
```

doc/manual/source/manual-page.rst
```diff
@@ -1062,6 +1067,11 @@ All values can be overridden via the command line options.
             CA may be away from a trust anchor certificate before it is
             rejected. If the option is missing, a default of 32 will be used.
 
+      enable-dnssec
```
AlexanderBand, line 1070: s/dnssec/bgpsec

partim:
This also fixes #579.

src/engine.rs
```diff
@@ -291,6 +290,7 @@ impl Engine {
         ))
     }
 
+    /*
```
DRiKE, line 293: Is this obsolete and can/should it be removed?

src/metrics.rs
```diff
@@ -342,9 +363,88 @@ impl ops::AddAssign for PublicationMetrics {
...
+    fn add_assign(&mut self, other: Self) {
+        self.add_assign(&other)
+    }
+}
+
+
 //------------ VrpMetrics ----------------------------------------------------
 
-/// Metrics regarding the generated VRP set.
+/// Individual metrics the generated payload.
```
DRiKE, line 447:
```suggestion
/// Individual metrics regarding the generated payload.
```

src/output.rs
```diff
@@ -321,56 +365,87 @@ where
...
                         self.formatter.delimiter(target)?;
                     }
-                    self.formatter.origin(origin, info, target)?;
-                    self.state = StreamState::Origin { index, first: false };
-                    break
+                    self.formatter.router_key(key, info, target)?;
+                    return Ok(true)
                 }
-                Ok(true)
+                (true, StreamState::Done)
```
DRiKE, line 439: Seems like StreamState::Done has no functional use (anymore) and can be removed from the match if we simply return here:
```suggestion
               return Ok(false)
```

src/output.rs
```diff
@@ -321,56 +365,87 @@ where
...
             }
-            StreamState::Done => Ok(false)
-        }
+            StreamState::Done => return Ok(false)
+        };
+        self.state = next;
+        Ok(res)
```
DRiKE, line 444: `res` is always true here, so perhaps it can be removed from the `let (res, next) = match self.state` at line 370 and be simplified here to:
```suggestion
        Ok(true)
```

src/payload.rs
```diff
@@ -129,6 +150,55 @@ impl<'a> ProcessPubPoint for PubPointProcessor<'a> {
...
+                    uri
+                );
+                return Ok(())
+            }
+        };
+        let id = cert.subject_key_identifier();
+        let key = cert.subject_public_key_info();
+        if !key.allow_router_cert() {
+            warn!(
+                "{}: router certifcate has invalid key algorithm.", uri
```
DRiKE, line 182:
```suggestion
                "{}: router certificate has invalid key algorithm.", uri
```

AlexanderBand:
As far as I can see, in the manpage you don't yet describe the `routerKeys` member in the `jsonext` output.

src/payload.rs
```diff
@@ -776,25 +904,51 @@ impl PayloadSnapshot {
...
         &self.payload
     }
 
     /// Returns an iterator over the route origins.
     pub fn origins(
         &self
-    ) -> impl Iterator<Item = (RouteOrigin, &OriginInfo)> + '_ {
-        self.payload().iter().filter_map(|item| {
+    ) -> impl Iterator<Item = (RouteOrigin, &PayloadInfo)> + '_ {
+        self.payload()[..self.router_key_index].iter().filter_map(|item| {
```
DRiKE, line 915: The `filter_` part seems unnecessary assuming all `item`s up until index `self.router_key_index` are of the `Payload::Origin` enum variant anyway. Could a `map` suffice, e.g.
```rust
.iter().map(|item| {
    (item.0.origin, &item.1)
}
```
If not mistaking, the final `None` should come from `iter()` anyway.

src/payload.rs
```diff
@@ -776,25 +904,51 @@ impl PayloadSnapshot {
...
     }
 
-    /// Converts a shared snapshot into a VRP iterator.
-    pub fn into_vrp_iter(self: Arc<Self>) -> SnapshotVrpIter {
-        SnapshotVrpIter::new(self)
+    /// Returns an interator over the router keys.
+    pub fn router_keys(
+        &self
+    ) -> impl Iterator<Item = (&RouterKey, &PayloadInfo)> + '_ {
+        self.payload()[self.router_key_index..].iter().filter_map(|item| {
```
DRiKE, line 927: See previous comment, `filter_map` might be unnecessary here as well.

src/payload.rs
```diff
@@ -1392,11 +1611,11 @@ impl<'a> Iterator for OriginInfoIter<'a> {
 }
 
 
-//------------ RoaInfo -------------------------------------------------------
+//------------ PublishInfo ---------------------------------------------------
 
-/// Information about the ROA an origin came from.
+/// Information about the published object payload came from.
```
DRiKE, line 1616:
```suggestion
/// Information about the published object a payload came from.
```

src/engine.rs
```diff
@@ -1883,9 +1884,9 @@ pub trait ProcessPubPoint: Sized + Send + Sync {
     /// The method is given both the URI and the certificate. If it
     /// returns an error, the entire processing run will be aborted.
     fn process_ee_cert(
```
partim, line 1886: Consider renaming to `process_router_cert`

src/payload.rs
```diff
@@ -1561,5 +1799,77 @@ mod test {
...
-}
 
+    #[test]
+    /// Tests the order of data in a snapshot.
+    ///
+    /// Creates a snapshot via a builder and then checks that the various
+    /// iterators produce correct output.
+    ///
+    /// This assumes that `RouteOrigin` and `RouterKey` order correctly.
+    fn snapshot_order() {
```
DRiKE, line 1810: Perhaps pedantic and highly unlikely, but it seems that if for some reason the `iter`s and `filter_map .. collect`s result in empty collections, the tests will pass. Simple counters in the for loops could be used to test we have seen indeed 4 origins and 3 router keys.

DRiKE (approved):

partim:
@DRiKE As discussed Friday, I have changed the way the combined set of data is stored in `PayloadSnapshot` to have separate vecs for each payload type. Because the RTR server needs an ‘iterator’ over `Payload`, they still hold that type but we can now panic with confidence if the type is wrong. As this required some re-wiring, it would be good if you could have a look at that particular commit again.

src/output.rs
partim: It actually is still necessary. The stream essentially is an iterator but instead of returning an item, it writes it to the target. The return value of the method is whether there was a value. Because the stream writes the footer if it runs out of items, it needs to return `true` after that and can only return `false` if it is called once more. `StreamState::Done` signals that that has to happen next time.

src/output.rs
partim: Oh, good. That makes the code formatting much less ugly ;)

src/payload.rs
partim: This has now all been rewritten. See comment below.

src/payload.rs
partim: There are asserts checking that `orgins_iter` and `keys_iter` are empty after we went over the collected vecs. Doesn’t that – together with unwrapping the next call during the for loop – ensure that the collected vecs have the exact same size as the reference vecs?

src/engine.rs
partim: This has been forked off into #701.

DRiKE (commented):
This separate vec approach is much cleaner and improves the readability of the code a lot.
One thing that's got me puzzled though: it seems the `next_with_info()`, and thus the `next()` on `SnapshotArcOriginsIter` and `SnapshotArcRouterKeysIter` never return a `None` anymore. That is a bit confusing for an iterator, but moreover, the matching on the `StreamState` for `OutputStream` in `fn write_next` (output.rs, line 371), how can that still work?

src/http/delta.rs
```diff
@@ -320,26 +318,28 @@ impl Iterator for SnapshotStream {
     type Item = Vec<u8>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let snapshot = self.snapshot.as_ref()?;
+        use rpki::rtr::server::PayloadSet;
```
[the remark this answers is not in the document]
partim: Didn’t want to pollute the entire module with the name. Probably doesn’t matter too much either way.

partim:
They do return `None` by short-circuiting if `get` returns `None` by way of a sneaky question mark:

```
    pub fn next_with_info(&mut self) -> Option<(RouteOrigin, &PayloadInfo)> {
        let res = self.snapshot.origins.get(self.next)?;
        // ...
     }
```

doc/manual/source/manual-page.rst
```diff
@@ -457,24 +462,41 @@ These can be requested by providing different commands on the command line.
...
+                  The object will have a *type* of *roa* if it was derived
+                  from a valid ROA object, *cer* if it was derived from
+                  a published router certificate, or *exception* if it was an
+                  assertion in a local exception file.
+
+                  For RPKI objects, *uri* provides the rsync URI of the ROA
+                  or router certificate, *validity* provides the validity of
+                  the ROA itself, and *chainValidity* the validity
+                  considering the validity of the certificates along the
+                  validation chain.
```
AlexanderBand, line 499: Were you going to leave these here or leave the elaborate description to: 
https://routinator.docs.nlnetlabs.nl/en/latest/output-formats.html

doc/manual/source/manual-page.rst
```diff
@@ -457,24 +462,41 @@ These can be requested by providing different commands on the command line.
                   format.
 
            jsonext
-                  The list is placed into a JSON object with a single element
-                  *roas* which contains an array of objects with four
+                  The list is placed into a JSON object with two members:
```
AlexanderBand, line 465: It also has a *metadata* member right?

doc/manual/source/manual-page.rst
partim: I figured we leave it here for now and move it out either before or after release.

doc/manual/source/manual-page.rst
partim: Ah yes. I added more text.

PR #698, opened 2022-01-27
Wrap RwLock and Mutex to avoid having to unwrap all the time.

partim:


src/utils/sync.rs
```diff
@@ -0,0 +1,90 @@
...
+#[derive(Debug, Default)]
+pub struct Mutex<T: ?Sized>(StdMutex<T>);
+
+impl<T> Mutex<T> {
+    /// Creates a new mutex in unlocked state.
+    pub fn new(t: T) -> Self {
+        Mutex(StdMutex::new(t))
+    }
+
+    /// Consumes the mutes, returning the underlying data.
```
DRiKE, line 65:
```suggestion
    /// Consumes the mutex, returning the underlying data.
```

DRiKE (approved):

PR #703, opened 2022-02-02
Switch to the vault package repository for CentOS 8 now that it is EOL.

ximon18:
This PR "fixes" the failing CentOS 8 packaging workflow ([see here](https://github.com/NLnetLabs/routinator/actions/runs/1778642751)) by telling CentOS 8 to use vault packages as there are no "current" packages now that it is EOL. This will presumably allow anyone still running CentOS 8 to install a Routinator package.

Note: This PR doesn't attempt to address the separate question of whether or not it makes sense to keep packaging for CentOS 8 or if we should package instead or additionally for Rocky Linux, for example (and whether or not we would need to also extend packages.nlnetlabs.nl with a separate RockyLinux repository as well).

A successful run of the packaging workflow using the changes in this PR can be seen [here](https://github.com/NLnetLabs/routinator/actions/runs/1786566863).

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -150,6 +150,11 @@ jobs:
             apt-get install -y build-essential jq lintian pkg-config
             ;;
           centos)
+            if [[ "${MATRIX_IMAGE}" == "centos:8" ]]; then
+              # allow CentOS 8 to continue working now that it is EOL
+              # see: https://stackoverflow.com/a/70930049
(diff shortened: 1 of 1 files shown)


ximon18:
Oops, I forgot to wait for the package test which is failing. Investigating...

ximon18:
> Oops, I forgot to wait for the package test which is failing. Investigating...

I resolved this by using the CentOS compatible AlmaLinux LXD image instead.

partim (approved):

PR #710, opened 2022-02-07
Enable TCP keep alive on RTR sockets if configured.

### LINKED ISSUE #320: Support for optional keep-alive in TCP transport
It would be great to have some functionality to optionally add support for a keep-alive to be sent in TCP, at a configurable interval.  

The use case being; somewhere in the path between RTR and router there is something tracking TCP sessions, and it sometimes kills them because its timers are not readily configurable and has a timeout interval < RTR refresh_interval.  RTR alone often isn't chatty enough to keep sessions alive.  I've seen this recently when routinator is a kubernetes pod and IPVS TCP timeouts default to 900s, whilst RFC8210 recommends a default of 3600s for refresh interval.

RFC8210 encourages the keep-alive behaviour;
```
   "To reduce exposure to dropped but non-terminated sessions, both
   caches and routers SHOULD enable keep-alives when available in the
   chosen transport protocol.”
```

In testing this behaviour with routers, I've implemented a stream proxy in NGINX and told it to set `so_keepalive` on the RTR listener socket.  In testing, both Arista and IOS-XR respond to the TCP keep-alive as expected, and this is enough to keep IPVS from eating the sessions, and maintaining recommended RTR refresh interval of 3600s.

partim:
Fixes #320. Again.

### DIFF
--- a/src/rtr.rs
+++ b/src/rtr.rs
@@ -1,6 +1,7 @@
 /// Support for the RPKI-to-Router Protocol.
 
 use std::io;
+use std::convert::TryFrom;
 use std::future::Future;
 use std::net::{SocketAddr, TcpListener as StdListener};
 use std::pin::Pin;
@@ -134,9 +135,12 @@ impl RtrStream {
     fn new(
         sock: TcpStream,
         tls: Option<&TlsAcceptor>,
-        _keepalive: Option<Duration>,
+        keepalive: Option<Duration>,
         server_metrics: SharedRtrServerMetrics,
     ) -> Result<Self, io::Error> {
+        if let Some(duration) = keepalive {
+            Self::set_keepalive(&sock, duration)?
+        }
         let metrics = Arc::new(RtrClientMetrics::new(sock.local_addr()?.ip()));
         let client_metrics = metrics.clone();
         tokio::spawn(async move {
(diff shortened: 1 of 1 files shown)


AlexanderBand (approved):

src/rtr.rs
```diff
@@ -147,6 +150,32 @@ impl RtrStream {
...
+        sock: &TcpStream, duration: Duration
+    ) -> Result<(), io::Error>{
+        use std::convert::TryFrom;
+        use std::os::unix::io::AsRawFd;
+        use nix::sys::socket::{setsockopt, sockopt};
+
+        (|fd, duration: Duration| {
+            setsockopt(
+                fd, sockopt::TcpKeepInterval,
+                &u32::try_from(duration.as_secs()).unwrap_or(u32::MAX)
```
DRiKE, line 165: Wouldn't it make more sense to use a realistic/common value instead of `u32::MAX` to fall back on? Because with an interval time that large, I suspect that the keepalive will not work in practice even though it is technically enabled.

fwiw:
```
$ cat /proc/sys/net/ipv4/tcp_keepalive_time
7200
```

DRiKE (approved):

PR #735, opened 2022-03-28
When dumping RRDP, don’t copy the hash at the beginning of objects.

### LINKED ISSUE #733: Objects from RRDP repositories are "invalid" in `dump` output
For objects in RRDP repositories, Routinator writes a 32-byte hash header on disk. The `dump` command copies these files as-is, so the RRDP repository objects in the dump output contain this same header.

partim:
Fixes #733.

### DIFF
--- a/src/collector/rrdp.rs
+++ b/src/collector/rrdp.rs
@@ -2012,7 +2012,7 @@ impl RepositoryObject {
             Ok(target) => target,
             Err(err) => {
                 error!(
-                    "Failed to create RRDP target file {}: {}",
+                    "Failed to create target RRDP file {}: {}",
                     target_path.display(), err
                 );
                 return Err(Failed)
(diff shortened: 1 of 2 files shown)


src/collector/rrdp.rs
```diff
@@ -1997,6 +1989,51 @@ impl RepositoryObject {
...
+    /// added hash.
+    pub fn dump(
+        source_path: &Path,
+        target_path: &Path,
+    ) -> Result<(), Failed> {
+        let mut source = match File::open(source_path) {
+            Ok(source) => source,
+            Err(err) => {
+                error!(
+                    "Failed to open source RRDP file {}: {}",
```
DRiKE, line 2005: Consistency with other error messages:
```suggestion
                    "Failed to open RRDP source file {}: {}",
```

DRiKE (approved):

src/collector/rrdp.rs
partim: I kinda like my own version better: This is really is about the source of the RRDP file to be dumped, not about the RRDP source of a file.

PR #743, opened 2022-04-04
Upgrade to syslog 6.

partim:
Somehow, the type for the pid on `Formatter3164` has changed to a `u32` while the pid returned by nix is a `i32`. Since `Formatter3164` comes with an impl of `Default` that mostly does the right thing, we switch to that and adjust as needed after.


partim:
Looks like this PR requires an upgrade of the MSRV so we have to leave it to 0.12.

PR #755, opened 2022-07-06
Adjust error handling to rpki-rs 0.15.

partim:


src/engine.rs
```diff
@@ -1424,25 +1431,26 @@ struct ValidPointManifest {
...
+                    "certificate has no CRL URI"
+                ).into())
             }
         };
 
         if *crl_uri != self.crl_uri {
-            warn!("{}: certifacte's CRL differs from manifest's.", uri);
-            return Err(ValidationError)
+            return Err(InspectionError::new(
+                "certifacte's CRL differs from manifest's"
```
DRiKE, line 1446:
```suggestion
                "certificate's CRL differs from manifest's"
```

src/engine.rs
```diff
@@ -1424,25 +1431,26 @@ struct ValidPointManifest {
...
+            ).into())
         }
 
         if self.crl.contains(cert.serial_number()) {
-            warn!("{}: certificate has been revoked.", uri);
-            return Err(ValidationError)
+            return Err(InspectionError::new(
+                "certifacte has been revoked"
```
DRiKE, line 1452:
```suggestion
                "certificate has been revoked"
```

src/engine.rs
```diff
@@ -1250,13 +1255,14 @@ impl<'a, P: ProcessRun> PubPoint<'a, P> {
             self.cert.cert(), self.run.validation.strict
         ) {
             Ok(cert) => cert,
-            Err(_) => {
-                warn!("{}: CA certificate failed to validate.", uri);
+            Err(err) => {
+                warn!("{}: {}.", uri, err);
```
DRiKE, line 1259: With both CA certs and router certs having file extension `.cer`, are we still able to determine which of the two types the error message describes without depending on the actual (contents of) `err`?

Would it make sense to add back 'CA' and 'router' in the `warn!`?

src/engine.rs
```diff
@@ -1307,14 +1313,15 @@ impl<'a, P: ProcessRun> PubPoint<'a, P> {
...
         manifest: &mut ValidPointManifest,
     ) -> Result<(), Failed> {
-        if cert.validate_router(
+        if let Err(err) = cert.validate_router(
             self.cert.cert(), self.run.validation.strict
-        ).is_err() {
-            warn!("{}: router certificate failed to validate.", uri);
+        ) {
+            warn!("{}: {}.", uri, err);
             manifest.metrics.invalid_certs += 1;
```
DRiKE, line 1320: see previous comment

DRiKE (approved):
Perhaps we should dive a bit deeper into where that unused `pub fn verify_hash` in store.rs comes from, but otherwise, only some nits.

src/engine.rs
partim: I believe the error messages contain the type of certificate if it is relevant. I think not including the type when it doesn’t matter might be clearer?

partim:
Looked into the unused verify hash function. Turns out that while we do (optionally) store the hash of a object with the object, we don’t need to check since we only ever store publication points that have been validated, so we know that the object in its stored form is on the manifest. We now retain the hash in the stored format for compatibility.

PR #758, opened 2022-08-02
Packaging fixes

ximon18:
This PR:
- Fixes the failure on `main` by disabling testing of the generated DEB package for the Debian Stretch platform, as the Debian Stretch LXC image is no longer available on images.linuxcontainers.org.
- Removes the need to use an old Rust version to build Routinator when generating RPMs by upgrading to latest `cargo-generate-rpm`.
- Fixes a minor issue where one of the step names still referred to DEB packaging even though it is also run for RPMs.
- Upgrades cargo-deb to 1.36.0, the newest version that both supports Rust edition 2021 (to support PR #760) **AND** compiles without `lzma` support (required with newer Cargo deb versions in order to build a valid package on Ubuntu Xenial). See cargo deb issues https://github.com/kornelski/cargo-deb/issues/12 and https://github.com/kornelski/cargo-deb/issues/43.

A successful run of the `pkg` workflow using this PR branch can be seen [here](https://github.com/NLnetLabs/routinator/actions/runs/2782340811).

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -191,7 +191,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ~/.cargo/bin/cargo-generate-rpm
-        key: ${{ matrix.image }}-cargo-generate-rpm-${{ env.CARGO_GENERATE_RPM_VER }}-${{ startsWith(${OS_REL}, "xenial")}}
+        key: ${{ matrix.image }}-cargo-generate-rpm-${{ env.CARGO_GENERATE_RPM_VER }}-${{ endsWith(matrix.image, "xenial")}}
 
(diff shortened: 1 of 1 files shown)


partim:
In order to switch to edition 2021, we also need a newer version of cargo-deb. Should we do this here while we are at it?

partim (approved):

ximon18:
@partim: GH Actions run https://github.com/NLnetLabs/routinator/runs/7727404734?check_suite_focus=true shows the latest commit to this branch working, complete with new enough cargo deb (>= 1.34.0) for Rust edition 2021 support (e.g. see [this log line](https://github.com/NLnetLabs/routinator/runs/7727404575?check_suite_focus=true#step:11:28)).

PR #759, opened 2022-08-02
Test cargo generate metadata overwrite feature

ximon18:
This DRAFT PR demonstrates that the [new metadata-overwrite feature](https://github.com/cat-in-136/cargo-generate-rpm/pull/36) being developed for `cargo-generate-rpm` can be used to avoid rewriting Cargo.toml during the packaging process in order to set the version correctly. It shouldn't be merged, rather if and when the upstream PR is merged we should return to this PR to update it to use the next release of `cargo-generate-rpm` that includes the new functionality.

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -98,7 +98,7 @@ jobs:
             target: 'aarch64-unknown-linux-musl'
     env:
       CARGO_DEB_VER: 1.28.0
-      CARGO_GENERATE_RPM_VER: 0.6.0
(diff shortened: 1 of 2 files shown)


ximon18:
I've updated the PR to use the upstream updated syntax. GH Actions `pkg` run is in progress here: https://github.com/NLnetLabs/routinator/actions/runs/2818889207

ximon18:
[GH Action run](https://github.com/NLnetLabs/routinator/actions/runs/2875561846) passed using release version 0.8.0 of cargo-generate-rpm.

PR #762, opened 2022-08-10
Enable building multi-architecture docker images

bladeswords:
Use QEMU to enable building containers for the ARM platform.  

This will increase the build time due to the emulation.

The testing step was removed as the load command did not work.

### DIFF
--- a/.github/workflows/pkg-docker.yml
+++ b/.github/workflows/pkg-docker.yml
@@ -21,7 +21,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v3
         with:
-          images: nlnetlabs/routinator
+          images: ${{ secrets.DOCKER_HUB_ID }}/routinator
           flavor: |
             latest=false
           tags: |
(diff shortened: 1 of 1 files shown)


bladeswords:
This addresses issue #763

.github/workflows/pkg-docker.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v3
         with:
-          images: nlnetlabs/routinator
+          images: ${{ secrets.DOCKER_HUB_ID }}/routinator
```
ximon18, line 24: FYI using the DOCKER_HUB_ID secret this isn't a good idea here as the Docker Hub ID can be that of a CI user, e.g. `nlnetlabsci`, rather than the organization name that is part of the image tag.

ximon18:
When I worked on the ARM cross compilation support for Krill I did things a little differently. I found that QEmu based compilation was exceedingly slow, it was quicker to cross compile outside of QEmu, and also that using the docker buildx support for building for multiple platforms at once on GH Actions doesn't scale because it tries to run them all in parallel on the same GH Actions runner host. As such I split things out and used GH Actions matrices to parallelize the building over multiple runners. In this particular case you only add one platform but for Krill we have it currently building for 4 platforms, I imagine we would do the same for Routinator too.

I'll discuss this internally, but for the reasons above and for consistency across projects my first thought is to use the same approach as I used for Krill. Hopefully I can also at some point start to extract common GH Actions workflow elements out to the NLnet Labs organization level and re-use them in multiple projects as well, e.g. this may also apply to `rtrtr`.

.github/workflows/pkg-docker.yml
bladeswords: Thanks!  This makes sense - I didn't think of that situation.  Great pickup.

ximon18:
Hi @bladeswords,

I've created PR #772 to extend our packaging workflow to build multi-arch Docker images and as part of testing it I pushed a multi-arch image to `ximoneighteen/routinator:unstable`, if you could verify that a basic sanity check of it works as expected that would be very appreciated.

Thanks,

Ximon

Issue #763, opened 2022-08-10
Build multi-architecture docker images


bladeswords:
It would be great if the ARM instruction set was supported in the docker builds published to dockerhub.  This would enable easy use routinator on small devices such as raspberry pi and also modern cloud such as AWS graviton instances.

ximon18:
Hi @bladeswords,

Thanks for the request, the description of potential use cases and most of all for a PR!

As it happens I just extended the Krill GitHub Actions workflows to publish multi-arch Docker images to Docker Hub for several ARM variants.

I'll take a look at your PR.

Ximon

bladeswords:
Amazing, if you can reuse what you did for Krill - keep it going.  It does seem like a better way forward over using QEMU.

PR #764, opened 2022-08-10
dockerfile to support multiarchitecture

bladeswords:
Bump alpine to latest version
adjust build path to support building on different architectures such as ARM.
Addresses #763

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # -- stage 0: some configuration
-ARG BASE_IMG=alpine:3.15
+ARG BASE_IMG=alpine:3.16
 
 # -- stage 1: build static routinator with musl libc for alpine
 FROM ${BASE_IMG} as build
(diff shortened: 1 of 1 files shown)


ximon18:
Thanks @bladeswords.

> Bump alpine to latest version

Is that related to the main purpose of the PR or something you did as well at the same time but unrelated?

Thanks,

Ximon

bladeswords:
> Is that related to the main purpose of the PR or something you did as well at the same time but unrelated?

I did that at the same time as removing the specific platform target.  Specifying the x86_64-alpine-linux-musl target prevented this from building on an arm system.  Removing the target allowed the compiler to select and build for the right platform.

BTW, I have never used rust before so I am very unfamiliar with the terminology and tooling - sorry.

ximon18:
> BTW, I have never used rust before so I am very unfamiliar with the terminology and tooling - sorry.

Absolutely no need to apologize, the contribution and input is extremely welcome!

ximon18:
> > Is that related to the main purpose of the PR or something you did as well at the same time but unrelated?
> 
> I did that at the same time as removing the specific platform target. Specifying the x86_64-alpine-linux-musl target prevented this from building on an arm system. Removing the target allowed the compiler to select and build for the right platform.
> 
> BTW, I have never used rust before so I am very unfamiliar with the terminology and tooling - sorry.

I've included the change to not pass `--target` in PR #764, thanks!

Issue #774, opened 2022-08-23
User 'routinator' already exists on RPM re-install


ximon18:
Platform: Fedora 36 x86_64

Having had the Routinator RPM installed and then uninstalled, a new install reports an error:

```
$ sudo dnf install ./routinator-0.12.0-dev-1.x86_64.rpm
Last metadata expiration check: 4:56:18 ago on Tue 23 Aug 2022 17:08:27 CEST.
Dependencies resolved.
==============================================================================================================================================================================================================================================
 Package                                                  Architecture                                         Version                                                       Repository                                                  Size
==============================================================================================================================================================================================================================================
Installing:
 routinator                                               x86_64                                               0.12.0-dev-1                                                  @commandline                                               2.7 M

Transaction Summary
==============================================================================================================================================================================================================================================
Install  1 Package

Total size: 2.7 M
Installed size: 8.6 M
Is this ok [y/N]: y
Downloading Packages:
Running transaction check
Transaction check succeeded.
Running transaction test
Transaction test succeeded.
Running transaction
  Preparing        :                                                                                                                                                                                                                      1/1 
  Installing       : routinator-0.12.0-dev-1.x86_64                                                                                                                                                                                       1/1 
  Running scriptlet: routinator-0.12.0-dev-1.x86_64                                                                                                                                                                                       1/1 
useradd: user 'routinator' already exists
warning: %post(routinator-0:0.12.0-dev-1.x86_64) scriptlet failed, exit status 9

Error in POSTIN scriptlet in rpm package routinator
  Verifying        : routinator-0.12.0-dev-1.x86_64                                                                                                                                                                                       1/1 

Installed:
  routinator-0.12.0-dev-1.x86_64                                                                                                                                                                                                              

Complete!
```

Notice the errors:
```
useradd: user 'routinator' already exists
warning: %post(routinator-0:0.12.0-dev-1.x86_64) scriptlet failed, exit status 9

Error in POSTIN scriptlet in rpm package routinator
```

ximon18:
If I recall correctly the guidance is to leave created user accounts behind as they may still own files, so we should accept an existing user account.

ximon18:
Fixed by PR #780.

Issue #773, opened 2022-08-23
RPM postuninst script fails to run


ximon18:
Platform: Fedora 36 x86_64

```
$ ls -la /usr/share/routinator/rpm/
total 12
drwxr-xr-x 1 root root   54 Aug 23 22:03 .
drwxr-xr-x 1 root root    6 Aug 23 22:02 ..
-rwxr-xr-x 1 root root 1073 Aug 23 17:53 postinst
-rwxr-xr-x 1 root root  378 Aug 23 17:53 postuninst
-rwxr-xr-x 1 root root  436 Aug 23 17:53 preuninst
```

Clearly the `postuninst` script exists.

```
$ sudo dnf remove routinator
Dependencies resolved.
==============================================================================================================================================================================================================================================
 Package                                                    Architecture                                           Version                                                      Repository                                               Size
==============================================================================================================================================================================================================================================
Removing:
 routinator                                                 x86_64                                                 0.12.0-dev-1                                                 @System                                                 8.6 M

Transaction Summary
==============================================================================================================================================================================================================================================
Remove  1 Package

Freed space: 8.6 M
Is this ok [y/N]: y
Running transaction check
Transaction check succeeded.
Running transaction test
Transaction test succeeded.
Running transaction
  Preparing        :                                                                                                                                                                                                                      1/1 
  Running scriptlet: routinator-0.12.0-dev-1.x86_64                                                                                                                                                                                       1/1 
  Erasing          : routinator-0.12.0-dev-1.x86_64                                                                                                                                                                                       1/1 
  Running scriptlet: routinator-0.12.0-dev-1.x86_64                                                                                                                                                                                       1/1 
/var/tmp/rpm-tmp.Q0i4Rw: line 1: /usr/share/routinator/rpm/postuninst: No such file or directory
warning: %postun(routinator-0:0.12.0-dev-1.x86_64) scriptlet failed, exit status 127

Error in POSTUN scriptlet in rpm package routinator
  Verifying        : routinator-0.12.0-dev-1.x86_64                                                                                                                                                                                       1/1 

Removed:
  routinator-0.12.0-dev-1.x86_64                                                                                                                                                                                                              

Complete!
```

Notice these errors:
```
/var/tmp/rpm-tmp.Q0i4Rw: line 1: /usr/share/routinator/rpm/postuninst: No such file or directory
warning: %postun(routinator-0:0.12.0-dev-1.x86_64) scriptlet failed, exit status 127

[...]

PR #776, opened 2022-08-24
RPM (un)install fixes for #773 and #774

ximon18:
This PR:
- Extends the packaging GitHub Actions workflow to attempt uninstallation and re-installation and to detect any occurrences of strings in the output of the install and uninstall processes that contain "err"(or), "warn"(ing) or "fail".
- Embeds RPM scriptlets in the RPM package directly so that on uninstall the script commands are available to run whereas previously it attempted to run a script stored on disk which had already been removed (#773).
- Doesn't try and create the `routinator` user on installation if the user already exists, which should be the case if Routinator was previously installed and uninstalled (as it is considered best practice to leave a package specific user account behind when uninstalling the package in case there are still files owned by that user) (#774).

The changes can be seen working in [this packaging workflow run](https://github.com/NLnetLabs/routinator/actions/runs/2921327622).

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -654,3 +654,25 @@ jobs:
 
         echo -e "\nROUTINATOR RPKI CACHE DIR (first 20 lines of ls output only):"
         sg lxd -c "lxc exec testcon -- ls -ltR /var/lib/routinator/rpki-cache/ | head -n 20"
+
+    - name: Test uninstall (without purge)
+      run: |
+        case ${OS_NAME} in
+          debian|ubuntu)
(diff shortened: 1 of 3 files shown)


ximon18:
This PR is obsoleted by PR #780. PR #780 switches to using a reusable workflow defined in the organization wide NLnetLabs/.github repository which includes the work done for this PR and more. PR #780 should be preferred over this PR.

ximon18:
This PR is obsoleted by PR #780. PR #780 switches to using a reusable workflow defined in the organization wide NLnetLabs/.github repository which includes the work done for this PR and more. PR #780 should be preferred over this PR.

PR #786, opened 2022-09-15
inode usage requirements

AlexanderBand:


doc/manual/source/installation.rst
```diff
@@ -4,10 +4,19 @@ Installation
...
+4GB of disk space for the application. 
+
+Please keep in mind that the RPKI consists of a very large number of small
+files. As a result, Routinator will use a large number of inodes. You should
+accommodate for at least 500,000 inodes, but one million will provide more
+breathing room. This will give you ample margin for the RPKI repositories to
+grow over time, as adoption increases. 
+
+.. Note:: A "No space left on device" error may be caused by running out of 
+          inodes instead of disk space. The ``du --inode`` command will list
```
partim, line 18: You might want to also mention `df -i` to show the amount of inodes available, used, and free.

partim: Actually, I think `df -i` is the more helpful command in this case. `du --inode` tells you how many inodes Routinator is currently using (if you run it in the right directory).

partim (approved):

PR #791, opened 2022-10-03
Update clap to version 4.

partim:


src/config.rs
```diff
@@ -848,8 +854,10 @@ impl Config {
         }
 
         // log_level
-        match (matches.occurrences_of("verbose"),
-                                            matches.occurrences_of("quiet")) {
+        match (
```
DRiKE, line 857: I don't think this is new in this PR per se, but this code doesn't make sense (anymore) with the `.conflicts_with` on the `verbose`/`quiet` args. Running the binary will not get past the arg parsing anyway.

src/config.rs
```diff
@@ -304,195 +304,203 @@ impl Config {
...
-        .arg(Arg::with_name("syslog-facility")
-             .long("syslog-facility")
-             .takes_value(true)
-             .default_value("daemon")
-             .help("Facility to use for syslog logging")
+        .arg(Arg::new("syslog-facility")
+            .long("syslog-facility")
+            .default_value("daemon")
+            .action(ArgAction::Set)
+            .help("Facility to use for syslog logging")
```
DRiKE, line 497: The default value is not mentioned in this help message, is that on purpose?

DRiKE (approved):

src/config.rs
partim: Hm, yeah. But the code is cute ;)

src/config.rs
partim: It is actually added by the help generator thingamajig:
```
      --syslog-facility <FACILITY>
          Facility to use for syslog logging [default: daemon]
```

PR #795, opened 2022-10-05
Convert command line parsing to clap’s derive API.

partim:
This PR converts command line parsing to a more safe implementation using clap’s derive API in most places.

Note that the init command has not been converted as it is going to be removed very soon.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -606,7 +606,7 @@ impl Config {
     #[allow(clippy::unnecessary_wraps)]
     fn apply_log_matches(
         &mut self,
-        matches: &ArgMatches,
+        args: &GlobalArgs,
         cur_dir: &Path,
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -305,204 +312,7 @@ impl Config {
     /// The function follows clap’s builder pattern: it takes an app,
```
DRiKE, line 312: update comment: builder pattern -> derive

src/config.rs
```diff
@@ -515,126 +325,8 @@ impl Config {
     ///
     /// It follows clap’s builder pattern and returns the app with all
```
DRiKE, line 326: update comment: builder pattern -> derive

DRiKE (approved):

PR #799, opened 2022-10-12
Add an rrdp-fallback option to configure fallback to rsync.

### LINKED ISSUE #730: Allow disabling rsync fallback.
Add a new option that allows users to choose whether and how to fall back to rsync. Possible options are (a) always, (b) never, or (c) only for unknown repositories.

partim:
This PR introduces a new command line and config option `rrdp-fallback` that decides when to fall back to rsync for a failed RRDP update. The available policies are `never` for never, `stale` for fallback after a certain time of falling the RRDP update, and `new` for fallback when RRDP has never succeeded. The default is `stale` which reflects the current behavior.

Fixes #730.

### DIFF
--- a/src/collector/base.rs
+++ b/src/collector/base.rs
@@ -9,7 +9,7 @@ use bytes::Bytes;
 use log::error;
 use rpki::repository::tal::TalUri;
 use rpki::uri;
-use crate::config::Config;
+use crate::config::{Config, FallbackPolicy};
 use crate::error::Failed;
 use crate::metrics::Metrics;
 use crate::engine::CaCert;
@@ -39,6 +39,9 @@ pub struct Collector {
     ///
     /// If this is `None`, use of rsync has been disabled entirely.
     rsync: Option<rsync::Collector>,
+
+    /// The policy for falling back from RRDP to rsync.
+    rrdp_fallback: FallbackPolicy,
 }
 
 impl Collector {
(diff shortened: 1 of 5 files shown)


AlexanderBand (approved):

src/collector/base.rs
```diff
@@ -184,15 +192,50 @@ impl<'a> Run<'a> {
...
+                        }
+                    }
+                    rrdp::LoadResult::Stale => {
+                        // Update failed and data is now stale. Only
+                        // "stale" wants us to fall back, "never" and "new"
+                        // want us to fail.
+                        if matches!(
+                            self.collector.rrdp_fallback,
+                            FallbackPolicy::Never | FallbackPolicy::New
+                        ) {
```
DRiKE, line 222:
```suggestion
                        if !matches!(
                            self.collector.rrdp_fallback,
                            FallbackPolicy::Stale
                        ) {
```

DRiKE: If ever another variant is added to FallbackPolicy, this will still be correct.

src/collector/rrdp.rs
```diff
@@ -288,7 +288,7 @@ pub struct Run<'a> {
     ///
     /// If there is some value for a repository, it is available and current.
```
DRiKE, line 289: comment outdated

src/collector/rrdp.rs
```diff
@@ -560,14 +579,14 @@ impl Repository {
     /// downloaded version that hasn’t expired yet.
```
DRiKE, line 579: outdated comment

DRiKE (approved):

PR #810, opened 2022-10-17
Allow limiting the prefix length of included VRPs.

### LINKED ISSUE #234: Allow setting maxium prefix length
Provide an option to filter prefixes over a certain prefix length. Note that this is different from max-length.

Question is, should we allow to configure this or just use /24 for v4 and /48 for v6? The latter is probably what everyone wants, anyway.

partim:
This PR adds two new command line and config file options, `limit-v4-len` and `limit-v6-len`, that allow limiting the prefix length of IPv4 or IPv6 prefixes, respectively, for which VRPs are to be included in the output data set. By default, all valid VRPs are included.

Fixes #234.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1672,7 +1672,7 @@ struct GlobalArgs {
     #[arg(
         long,
         value_name = "LENGTH",
-        value_parser = clap::value_parser!(u8).range(..32)
+        value_parser = clap::value_parser!(u8).range(..=32)
     )]
     limit_v4_len: Option<u8>,
 
@@ -1680,7 +1680,7 @@ struct GlobalArgs {
     #[arg(
         long,
         value_name = "LENGTH",
-        value_parser = clap::value_parser!(u8).range(..128)
+        value_parser = clap::value_parser!(u8).range(..=128)
     )]
     limit_v6_len: Option<u8>,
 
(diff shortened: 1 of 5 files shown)


AlexanderBand (approved):

src/config.rs
```diff
@@ -1641,6 +1668,22 @@ struct GlobalArgs {
     #[arg(long, value_name = "POLICY")]
     unknown_objects: Option<FilterPolicy>,
 
+    /// Maximum length of IPv4 prefixes included in output
+    #[arg(
+        long,
+        value_name = "LENGTH",
+        value_parser = clap::value_parser!(u8).range(..32)
```
DRiKE, line 1675: This will error if '32' is passed, as the range is not inclusive. Is that what we want? The `take_limited_u8` will allow 32 in the config file, so at the very least things are inconsistent.

src/config.rs
```diff
@@ -1641,6 +1668,22 @@ struct GlobalArgs {
...
+    )]
+    limit_v4_len: Option<u8>,
+
+    /// Maximum length of IP64 prefixes included in output
+    #[arg(
+        long,
+        value_name = "LENGTH",
+        value_parser = clap::value_parser!(u8).range(..128)
```
DRiKE, line 1683: Same as above

DRiKE (changes requested):

src/config.rs
partim: Uh, nice find!

DRiKE (approved):

PR #811, opened 2022-10-19
Fix RTR client metrics.

partim:
This PR fixes two issues with RTR client metrics: For one, we accidentally used the local address to identify the client instead of the peer address. And second, the collapsing of multiple client metrics objects was broken.


DRiKE (approved):
Additional comments on some of the if-let-else blocks within the `if collapse` block might be nice. The big comment that's already there is helpful but going over the code mentally is a bit error prone, some brief comments might serve in correctly following all the steps.

partim:
You are right: Done.

PR #816, opened 2022-11-02
Upgrade reusable packaging workflow

ximon18:
This PR upgrades the `pkg` workflow to use the latest version of the NLnet Labs [Ploutos](https://github.com/NLnetLabs/ploutos/) reusable packaging workflow. There have been a number of breaking releases of Ploutos fixing bugs, improving performance and simplifying its use:

- [v2](https://github.com/NLnetLabs/.github/releases/v2.0.0) checks for RPM scriptlet errors, eliminates 23 annoying Github workflow run warnings (caused by deprecated actions and the coming upgrade by GH to NodeJS 16) and introduces (optional) support for systemd unit management in RPM scriptlets. This PR adopts the new RPM systemd unit management behaviour. A check of the new managed behaviour vs the old manual behaviour shows that the only difference is use of newer systemctl command syntax on systems that support it, e.g. use of `--now` to achieve in one command what requires two commands on older systemd versions. The new systemd managed script fragments also include an explicit `systemd-reload`.

- [v3](https://github.com/NLnetLabs/.github/releases/v3.0.0) simplified usage by folding the `xxx_rules_path` functionality into the `xxx_rules` workflow inputs., fixes an issue with `minimal` profile selection on older O/S's, and removes the need for manual specification of cross-compilation targets (these are now determined automatically from the other workflow inputs). This PR updates the `pkg` workflow to follow the `xxx_rules` changes.

- [v4](https://github.com/NLnetLabs/.github/releases/v4.0.0) of the packaging workflow enables cross-compilation to take place in parallel with non-cross compilation thereby speeding up the workflow. In the case of Routinator this seems to reduce the time from ~16m to ~8m, about a 50% reduction!

- [v5](https://github.com/NLnetLabs/ploutos/releases/tag/v5.0.0) simplified usage by removing the need to specify redundant DEB maintainer name/email and package testing rule inputs, and fixes issues with slow `yum` operations and broken DEB changelog generation.

This PR:
- Adjusts the inputs passed in `pkg.yml` to Ploutou to match the breaking changes made to Ploutos.
  - Remove no longer supported or needed `deb_maintainer` Ploutos input.
  - Remove no longer needed `cross_build_rules` Ploutos input and associated `pkg/rules/cross-targets.yml` file, Ploutos will determine cross-compilation targets itself based on the `package_build_rules` input.
  - Remove no longer needed `package_test_rules` Ploutos input and associated `pkg/rules/packages-to-test.yml` file.
  - Updates `pkg/rules/packages-to-build.yml` to include the `mode` key used by the Ploutos `pkg-test` job now that `packages-to-test.yml` no longer defines it.
  - Replaces manual invocation of systemd commands in `pkg/rpm/scriptlets.toml` with the `#RPM_SYSTEMD_MACROS#` placeholder instead which Ploutos replaces at package building time with appropriate systemd commands (just as cargo-deb does for DEBs).
- Re-orders and groups related Ploutos workflow inputs in `pkg.yml`.
- Removes the unused minimal-cross cargo-deb variant from `Cargo.toml` and the related `pkg/common/routinator-minimal-cross.routinator.service` file.

A test run of the `pkg` workflow in this PR branch can be seen here: https://github.com/NLnetLabs/routinator/actions/runs/3591659903.


ximon18:
Warning! See: https://github.com/NLnetLabs/.github/issues/55

ximon18:
> Warning! See: [NLnetLabs/.github#55](https://github.com/NLnetLabs/.github/issues/55) Update: The identified issue doesn't currently affect Routinator but may do so if we upgrade to [v3.0.1](https://github.com/NLnetLabs/.github/releases/v3.0.1) of the packaging workflow, at which point we will need PR #818.

We don't actually need PR #818 which fixes the minimal-cross profile because the minimal-cross profile is never used. It only applies in the case of BOTH cross-compilation AND targetting an "old" O/S release (xenial, bionic or stretch) but all of our cross-compilation targets for Routinator are modern O/S releases (bullseye and buster). So instead we can just remove the minimal-cross profile completely.

ximon18:
The `pkg` workflow is running against this updated PR branch here: https://github.com/NLnetLabs/routinator/actions/runs/3417730321

partim (approved):

PR #854, opened 2023-04-26
Refer to (standalone) EE certs as router certs.

### LINKED ISSUE #701: Internally rename standalone EE certs to router certs.
This is to avoid confusion between EE certs embedded in signed objects and standalone EE certs that, at least for now, have to be router certificates.

partim:
This PR changes renames functions and attributes etc. that refer to standalone end-entity certificates to refer to them as ‘router certs’ to better distinguish them from EE certificates that are part of signed objects.

The only external change is in the HTTP status API where the `validEECerts` member has been renamed to `validRouterCerts` but the old name has been retained for now.

This is a breaking change.

Fixes #701.


DRiKE (approved):
Note that the docs on Prometheus metrics also (still) mention 'End Entity certificates' when it comes to `router_cert`. Which is not incorrect, but possibly confusing as well.

partim:
Good point. Fixed.

PR #859, opened 2023-05-03
Rewrite logging.

### LINKED ISSUE #468: Look into providing log file rotation.

### LINKED ISSUE #732: Simplify logging setup.
* Implement our own log wrapper instead of relying on `log-reroute`.
* Investigate whether using something simpler than `fern` makes sense.

partim:
This PR implements all log handling with the exception of actual syslog in Routinator itself. It also implements support for log rotation when logging into files by re-opening the log file when receiving SIGUSR2.

Error handling for logging is now such that if trying to log to file or syslog fails, Routinator will exit. It will also exit if it receives SIGUSR2 and can’t open the log file.

The motivation for this is that the log is used by many people to determine issues with the RPKI repositories, so silently not having logs seems bad. Also, not being able to log is a good indication for bigger problems to come.

Fixes #468. Fixes #732.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -2,12 +2,14 @@
 
 use std::{fs, io};
 use std::future::Future;
+use std::io::Write;
 use std::net::TcpListener;
 use std::path::Path;
 use std::sync::mpsc;
 use bytes::Bytes;
 use chrono::{DateTime, Utc};
 use log::{error, LevelFilter};
+use once_cell::sync::OnceCell;
 use tokio::runtime::Runtime;
 use crate::config::{Config, LogTarget};
 use crate::error::Failed;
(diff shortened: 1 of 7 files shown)


src/process.rs
```diff
@@ -310,46 +198,411 @@ impl Process {
...
+    }
+
+    /// Tries logging.
+    fn log(&mut self, record: &log::Record) -> Result<(), io::Error> {
+        match record.level() {
+            log::Level::Error => self.0.err(record.args()),
+            log::Level::Warn => self.0.warning(record.args()),
+            log::Level::Info => self.0.info(record.args()),
+            log::Level::Debug => self.0.debug(record.args()),
+            log::Level::Trace => self.0.debug(record.args()),
```
DRiKE, line 487:
```suggestion
            // Syslog has no trace level.
            log::Level::Trace => self.0.debug(record.args()),
```

src/process.rs
```diff
@@ -310,46 +198,411 @@ impl Process {
...
+            )
         };
-        (tx, res)
+        (queue, res)
     }
 
     pub fn start(&self) {
-        self.current.write().1 = Utc::now();
+        let new_string = format!(
+            "Log from validation run started at {}\n\n", Utc::now()
```
DRiKE, line 594: For the file/stderr logging, local time is used, not UTC. Shouldn't we use the same for both?
(Perhaps even make this configurable? Although on the other hand, I guess any serious logging happens using syslog anyway..)

src/process.rs
```diff
@@ -310,46 +198,411 @@ impl Process {
...
+            logger.flush()
+        }
+    }
+}
+
+
 //------------ LogOutput -----------------------------------------------------
 
 #[derive(Debug)]
 pub struct LogOutput {
```
DRiKE, line 575: So this is only used to optionally serve logging over HTTP, correct?
(I guess what I'm saying is, this part could benefit from some docstrings)

DRiKE (changes requested):
Nothing super interesting, but I found some comments regarding the log levels confusing. See proposed texts, feel free to alter/ignore.

PR #863, opened 2023-05-19
Add a notification endpoint to the json-delta protocol.

### LINKED ISSUE #723: Add a notification websocket to the JSON delta.
Currently, the `/json-delta` endpoint allows you to access differences in the data set in the same way as RTR via HTTP and JSON. However, this currently misses a way be notified when the data changes similar to the Serial Notify PDU in RTR. Add that using a websocket.

partim:
The PR adds a new endpoint `/json-delta/notify` that waits with responding until new data is available and then returns a JSON object with the session ID and serial number.

This seems to be a better solution than a websocket as it doesn’t require new dependencies.

Fixes #723.

### DIFF
--- a/doc/manual/source/manual-page.rst
+++ b/doc/manual/source/manual-page.rst
@@ -1378,6 +1378,11 @@ The service only supports GET requests with the following paths:
       provided session and serial. If *reset* is *true*, the *withdrawn*
       member is not present.
 
+/json-delta/notify
+      Waits with a response until a new update is available and then returns
+      a JSON object with two members *session* and *serial* which contain the
+      session ID and serial number of the updated data set.
+
 In addition, the current set of VRPs is available for each output format at a
 path with the same name as the output format. E.g., the CSV output is
 available at ``/csv``.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.3.0"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e579a7752471abc2a8268df8b20005e3eadd975f585398f17efcfd8d4927371"
+checksum = "0ca84f3628370c59db74ee214b3263d58f9aadd9b4fe7e711fd87dc452b7f163"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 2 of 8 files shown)


partim:
Hang on, I just realized there should be an option for a client to provide their session/serial and have the notify endpoint return immediately if there is different data. This allows a client to simply get notify again if their current get times out.

doc/manual/source/manual-page.rst
```diff
@@ -1378,6 +1378,16 @@ The service only supports GET requests with the following paths:
       provided session and serial. If *reset* is *true*, the *withdrawn*
       member is not present.
 
+/json-delta/notify, /json-delta/notify?session=session&serial=serial
+      Waits with a response until a new update is available and then returns
+      a JSON object with two members *session* and *serial* which contain the
+      session ID and serial number of the updated data set.
+
+      If the *session* and *serial* query parameters are provided, the JSON
+      object is returned immediately if the session ID and serial number of
```
DRiKE, line 1387: Presumably there should be a response if any of the two is different?
```suggestion
      object is returned immediately if the session ID or serial number of
```

doc/manual/source/manual-page.rst
```diff
@@ -1378,6 +1378,16 @@ The service only supports GET requests with the following paths:
...
+      Waits with a response until a new update is available and then returns
+      a JSON object with two members *session* and *serial* which contain the
+      session ID and serial number of the updated data set.
+
+      If the *session* and *serial* query parameters are provided, the JSON
+      object is returned immediately if the session ID or serial number of
+      the current data set differ from the provided values and only waits
+      for an update if they are identical.
+
 In addition, the current set of VRPs is available for each output format at a
```
density215, line 1391: > If the session and serial query parameters are provided,
What happens if I don't provide them? Also, don't have man pages have a syntax convention to indicate that query parms are optional?

density215: Maybe also indicate where the user could get the current session and serial info from? Or is that obvious from the workflow of using routinator?

density215 (approved):

doc/manual/source/manual-page.rst
partim: Maybe it would be better – and more consistent – to also return immediately if session/serial are missing? I.e., only wait if the current session/serial are provided?

partim: Ack. Change incoming.

AlexanderBand (approved):

PR #866, opened 2023-05-26
Allow excluding certain types in payload output.

### LINKED ISSUE #852: Endpoint for ASPAs
Create a dedicated endpoint to list ASPAs, e.g. `/aspa`, listing them in CSV format.

partim:
This PR add support for excluding certain payload types from output both in the `vrps` command and the HTTP server. With this in place, the PR also adds router key and ASPA to the standard `json` output format.

For the `vrps` command, new command line options `--no-route-origins`, `--no-router-keys`, and `--no-aspas` have been added. For the payload output endpoints of the HTTP server, a new `excludes` query parameter can be used to exclude `routeOrigins`, `routerKeys`, or `aspas`. 

These flags are available even for formats that don’t actually support outputting all these payload types.

When router keys or aspa has not been disabled, those payload types are now properly excluded rather than adding empty sections for them.

Fixes #852.

### DIFF
--- a/doc/manual/source/http-service.rst
+++ b/doc/manual/source/http-service.rst
@@ -52,7 +52,7 @@ This will produce the following output:
     }
 
 The query parameter ``exclude`` can be used to exclude certain payload types
-from the response. The values ``origins``, ``routerKeys``, and ``aspa``
+from the response. The values ``routeOrigins``, ``routerKeys``, and ``aspas``
 disable inclusion of route origins, router keys, and ASPAs, respectively. The
 values can either be given in separate ``exclude`` parameters or included in
 one separated by commas.
--- a/doc/manual/source/http-service.rst
+++ b/doc/manual/source/http-service.rst
@@ -52,7 +52,7 @@ This will produce the following output:
     }
 
 The query parameter ``exclude`` can be used to exclude certain payload types
-from the response. The values ``origins``, ``routerkeys``, and ``aspa``
+from the response. The values ``origins``, ``routerKeys``, and ``aspa``
 disable inclusion of route origins, router keys, and ASPAs, respectively. The
 values can either be given in separate ``exclude`` parameters or included in
 one separated by commas.
--- a/doc/manual/source/http-service.rst
+++ b/doc/manual/source/http-service.rst
@@ -51,6 +51,15 @@ This will produce the following output:
       ]
     }
 
+The query parameter ``exclude`` can be used to exclude certain payload types
+from the response. The values ``origins``, ``routerkeys``, and ``aspa``
+disable inclusion of route origins, router keys, and ASPAs, respectively. The
+values can either be given in separate ``exclude`` parameters or included in
+one separated by commas.
+
+.. versionadded:: 0.13.0
+   Allow excluding specific data from the output
+
 More Specific Prefixes
 """"""""""""""""""""""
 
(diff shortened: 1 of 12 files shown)


partim:
Come to think of it, now that the object members for excluded payload types disappear completely, we can add router keys and ASPA to regular JSON – they will not be present unless router keys and ASPA is enabled explicitly.

doc/manual/source/manual-page.rst
```diff
@@ -478,16 +495,23 @@ These can be requested by providing different commands on the command line.
...
+                  RPKI Validator except for different naming of the trust
+                  anchor.
+                  Routinator uses the name of the TAL file without the
+                  extension *.tal* whereas the RIPE NCC Validator has a
+                  dedicated name for each.
+
            jsonext
-                  The list is placed into a JSON object with three members:
-                  *roas* contains the validated route origin
+                  The list is placed into a JSON object with upt to four
```
AlexanderBand, line 507:
```suggestion
                  The list is placed into a JSON object with up to four
```

doc/manual/source/manual-page.rst
```diff
@@ -1391,6 +1420,12 @@ In addition, the query parameter ``include=more-specifics`` will cause the
 inclusion of VRPs for more specific prefixes of prefixes given via
 ``select-prefix``.
 
+Finally, the query parameters ``exclude`` can be used to exclude certain
```
AlexanderBand, line 1423:
```suggestion
Finally, the query parameter ``exclude`` can be used to exclude certain
```

AlexanderBand (changes requested):

doc/manual/source/manual-page.rst
```diff
@@ -458,18 +458,35 @@ These can be requested by providing different commands on the command line.
                   be in one single output file.
 
            json
-                  The list is placed into a JSON object with a single element
-                  *roas* which contains an array of objects with four
+                  The list is placed into a JSON object with up to four
+                  members: *roas* contains the validated route origin
+                  authorizations, *routerKeys* contains the validated 
+                  BGPsec router keys, *aspas* contains the validated
```
AlexanderBand, line 464: The `routerKeys` and `aspas` members are in the `json` format too, not just `jsonext`? I'm not seeing it in the output with this branch (yet).

src/output.rs
```diff
@@ -392,14 +290,188 @@ impl SelectResource {
...
+                    match value {
+                        "more-specifics" => selection.more_specifics = true,
+                        _ => { }
+                    }
+                }
+            }
+            else if key == "exclude" {
+                for value in value.split(',') {
+                    match value {
+                        "origins" => self.route_origins = false,
```
DRiKE, line 380: This is inconsistent with the command line switch `--no-route-origins`.

doc/manual/source/http-service.rst
```diff
@@ -51,6 +51,15 @@ This will produce the following output:
       ]
     }
 
+The query parameter ``exclude`` can be used to exclude certain payload types
+from the response. The values ``origins``, ``routerkeys``, and ``aspa``
```
DRiKE, line 55: `origins` here inconsistent with `--no-route-origins` on command line.

DRiKE (approved):

src/output.rs
partim: I was about to argue, but yeah, you are right.

PR #881, opened 2023-07-10
Package for Debian Bookworm

ximon18:
Successful test run: https://github.com/NLnetLabs/routinator/actions/runs/5511762595/jobs/10047867140#step:16:29


partim:
Do we want to keep an issue around (or even a draft PR) to enable the upgrade test again after 0.13 is released?

partim (approved):

PR #886, opened 2023-08-18
Store RRDP data in one file per repository.

partim:
This PR implements a very simple archive that can be used to store, update, and access the objects published by an RRDP repository in a single file.


src/utils/archive.rs
```diff
@@ -0,0 +1,1591 @@
...
+        }
+
+        // We are not in the chain at all???
+        Err(ArchiveError::Corrupt)
+    }
+
+    /// Finds the start of the object with the given name.
+    ///
+    /// The returned position is right after the name.
+    fn find(
```
timbru, line 426: So, there is a sequential scan whenever `find` is called, right? While this works, I wonder if a table of names to starting index would be useful (of course, it would have to be a bit smart about growth).

timbru (approved):
By and large looks good to me. I have (non-blocking) questions:

- would you consider keeping a growable index table to avoid sequential reads for `find`
- would you consider putting this in its own tiny crate so that it could be used by the `kvx` crate as well?

src/utils/archive.rs
partim: Because this is a hash map, the number of entries in a chain of each bucket should be really small. This can be optimized by choosing a good number of buckets. In most cases the file will be memory mapped, too, so there shouldn’t be actual reads either.

I _think_ this approach will be good enough, but we’ll have to see in practice.

partim:
Regarding re-using it in _kvx,_ this has been designed specifically for RRDP, not sure if it would also fit more generic use-cases. For instances, it can’t deal very well with growing objects. Not a problem with RRDP, but might be with other uses.

src/engine.rs
```diff
@@ -276,6 +276,18 @@ impl Engine {
...
+    /// This goes over the stored data and deletes what looks broken. It
+    /// should be called before retrying a failed restartable run.
+    pub fn sanitize(&self) -> Result<(), Fatal> {
+        self.store.sanitize()?;
+        if let Some(collector) = self.collector.as_ref() {
+            collector.sanitize()?;
+        }
+        Ok(())
+    }
+
```
density215, line 290: This still doesn't do anything, right? The .sanitize() method that is called doesn't do anything

density215 (approved):
Just a few remarks/suggestions.

src/utils/archive.rs
```diff
@@ -0,0 +1,1828 @@
...
+    /// The size of a single bucket.
+    ///
+    /// This is equal to the size of the integer type we are using for archive
+    /// positions, i.e., `u64`.
+    const BUCKET_SIZE: usize = mem::size_of::<u64>();
+
+    /// Returns the hash value for a given name.
+    ///
+    /// The returned value will already be taken module the number of buckets.
+    fn hash_name(&self, name: &[u8]) -> u64 {
```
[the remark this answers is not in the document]
partim: Yes, modulo. I extended the comment to say that this is really the bucket index, not just the hash.

(I don’t want to rename the method, though, because I can’t think of a good name …)

src/utils/archive.rs
```diff
@@ -0,0 +1,1828 @@
...
+}
+
+impl ArchiveMeta {
+    /// Creates a new value.
+    ///
+    /// This uses a random hash key and the given bucket number.
+    fn new(bucket_count: usize) -> Self {
+        ArchiveMeta {
+            hash_key: rand::random(),
+            bucket_count,
```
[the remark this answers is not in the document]
partim: I think that’s what you are supposed to do when initializing a SipHash. It _probably_ doesn’t matter, but why not do it properly just in case.

src/engine.rs
partim: The RRDP collector’s sanitize method checks for the consistency of all its archives.

PR #909, opened 2023-11-24
Accept EC PRIVATE KEY items in PEM files.

### LINKED ISSUE #908: Header/footer lines of EC TLS key files
This is related to https://github.com/NLnetLabs/routinator/issues/825

Routinator does not accept key files with header and footer lines like the following:
```
 -----BEGIN EC PRIVATE KEY-----
 -----END EC PRIVATE KEY-----
```

This results in
`[ERROR] TLS key file <keyfile> does not contain any usable keys`

Removing `EC` does the trick.

partim:
This PR adds support for private keys marked as “EC PRIVATE KEY“ in the PEM files for TLS server configuration.

Fixes #908.


src/utils/tls.rs
```diff
@@ -96,7 +96,7 @@ fn read_key(key_path: &Path) -> Result<PrivateKey, ExitError> {
         })?;
 
         let bits = match item {
-            RSAKey(bits) | PKCS8Key(bits) => bits,
+            RSAKey(bits) | PKCS8Key(bits) | ECKey(bits) => bits,
```
ximon18, line 99: Is this tested anywhere?

partim: Not currently, no.

ximon18: I tested it locally like so:

```
$ openssl ecparam -name prime256v1 -genkey -noout -out /tmp/ec-private-key.pem
$ openssl ec -in /tmp/ec-private-key.pem -pubout -out /tmp/ec-public-key.pem
$ openssl req -new -x509 -key /tmp/ec-private-key.pem -out /tmp/ec-cert.pem -days 360
$ routinator -vvv server --http-tls 127.0.0.1:8443 --http-tls-cert /tmp/ec-cert.pem --http-tls-key /tmp/ec-private-key.pem 
```

With Routinator 0.13.0 that resulted in an error:

```
[ERROR] TLS key file '/tmp/ec-private-key.pem' does not contain any usable keys.
```

With this PR there was no error and `openssl` shows the EC key in use:

```
$ true | openssl s_client -connect localhost:8443 2>&1 | openssl x509 -noout -text | grep Algorithm
        Signature Algorithm: ecdsa-with-SHA256
            Public Key Algorithm: id-ecPublicKey
    Signature Algorithm: ecdsa-with-SHA256
```

ximon18 (approved):

PR #910, opened 2023-11-29
Exclude the manual from the cargo package.

partim:
This PR excludes the `doc/manual` directory from the files included with the package uploaded to crates.io when publishing.


AlexanderBand (approved):

ximon18:
LGTM, but why not also exclude things like the `Dockerfile` and the RtD config?

PR #912, opened 2023-12-01
Replace `num_cpus` with `std::thread::available_parallelism`.

### LINKED ISSUE #902: Use std lib instead of num_cpus
Consider whether it might be possible to replace the dependency on the `num_cpus` crate by using the `std::thread::available_parallelism` functionality instead thus removing a 3rd party dependency.

partim:
This PR removes the dependency on the num_cpus crate by using the standard library’s `std::thread::available_parallelism` instead.

Fixes #902.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1150,7 +1150,6 @@ dependencies = [
  "listenfd",
  "log",
  "nix",
- "num_cpus",
  "pin-project-lite",
  "rand",
  "reqwest",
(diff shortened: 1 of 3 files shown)


ximon18:
LGTM, the fns called by both the old and the new code seem to query for some number of "logical" CPUs so that's consistent. One minor point: I did wonder whether it would make sense to extract this repeated code fragment into a helper fn:

```
available_parallelism().map(|x| x.get()).unwrap_or(1)
```

partim:
Sounds like a good idea, indeed. Done!

PR #916, opened 2024-01-10
Make dumping an empty cache succeed.

### LINKED ISSUE #914: routinator dump not working
I am running Routinator 0.13.0 (OS Ubuntu 20.04) and when I try to dump the routinator data I get the following error. The software creates the folder `cache`, creates the folder of `ta`-s inside of it but fails to create everything else. Even if I try to manually add `store/repositories.json` it won't recognize it.
```

$ routinator dump --output cache
[ERROR] Fatal: failed to write file /../cache/store/repositories.json: No such file or directory (os error 2)
[ERROR] Fatal error. Exiting.
```
Am I using the command wrong?

partim:
This PR fixes a number of error that can happen during `dump` when the cache is missing certain directories.

Fixes  #914.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,11 @@ on:
   push:
     branches: 
       - main
+      - series-*
   pull_request:
     branches: 
       - main
+      - series-*
 jobs:
   test:
     name: test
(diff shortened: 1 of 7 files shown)


partim:
This PR is for the 0.13 series. Once approved, it needs to be forked for main before merging.

DRiKE (approved):

PR #917, opened 2024-01-10
Print a more meaningful message when decoding an object fails.

partim:
This PR changes the error message printed when decoding any of the RPKI objects fails from whatever wondrous thing the decoder says to a generic message stating that decoding an object of a certain expected type failed.


partim:
The decoder does indeed not provide any useful information. This is a limitation of the current error handling model in both _bcder_ and _rpki-rs._ It can (and should) be fixed at some point. This PR really only provides a work-around for showing `missing further values (at position 2)` to users as the only error reason.

PR #918, opened 2024-01-10
Make dumping an empty cache succeed.

partim:
This PR fixes a number of error that can happen during dump when the cache is missing certain directories.

This is a clone of #916 applied to current main. It differs slightly in that it cleans up the changes to the `utils::fatal` module.

### DIFF
--- a/src/utils/fatal.rs
+++ b/src/utils/fatal.rs
@@ -318,12 +318,12 @@ pub fn write_file(path: &Path, contents: &[u8]) -> Result<(), Failed> {
 
 //------------ copy_dir_all --------------------------------------------------
 
-/// Copies the content of a directory.
+/// Copies the content of a directory if it exists.
 ///
 /// Creates the target directory with `create_dir_all`.  Errors out if
 /// anything goes wrong.
 ///
-/// If the source director does not exist, does nothing.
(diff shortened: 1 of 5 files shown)


src/utils/fatal.rs
```diff
@@ -322,9 +322,17 @@ pub fn write_file(path: &Path, contents: &[u8]) -> Result<(), Failed> {
 ///
 /// Creates the target directory with `create_dir_all`.  Errors out if
 /// anything goes wrong.
-pub fn copy_dir_all(source: &Path, target: &Path) -> Result<(), Failed> {
+///
+/// If the source director does not exist, does nothing.
```
ximon18, line 326: s/director/directory/

Other than that, I find this a bit confusing. It has been explicitly renamed to say it operates on an +_existing_+ directory, yet checks if it exists and succeeds without doing anything if it doesn't exist. That violates my expectations based on the name.

ximon18 (approved):

src/utils/fatal.rs
partim: The intention was to have a function that doesn’t error if something doesn’t exist. I adjusted the doc string slightly.

PR #920, opened 2024-01-11
Directly log rsync error output.

partim:
This PR changes the rsync collector to directly log all stderr output from the rsync command rather than collecting it and then blasting it out all at once which can cause issues with the syslog daemon on some systems.

### DIFF
--- a/src/collector/rsync.rs
+++ b/src/collector/rsync.rs
@@ -549,20 +549,7 @@ impl RsyncCommand {
                     if let Some(mut pipe) = stderr_pipe {
                         let mut line = Vec::new();
                         while pipe.read_until(b'\n', &mut line).await? != 0 {
-                            // Drop the delimiter if there is one, don’t
-                            // log an empty line.
-                            let len = line.len();
-                            if len > 0 && line[len - 1] == b'\n' {
-                                let _ = line.pop();
-                            }
-                            if line.is_empty() {
-                                continue
-                            }
-                            warn!(
-                                "{}: {}",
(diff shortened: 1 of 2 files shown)


src/collector/rsync.rs
```diff
@@ -545,7 +547,23 @@ impl RsyncCommand {
                 },
                 async {
                     if let Some(mut pipe) = stderr_pipe {
-                        tokio::io::copy(&mut pipe, &mut stderr).await?;
+                        let mut line = Vec::new();
+                        while pipe.read_until(b'\n', &mut line).await? != 0 {
+                            // Drop the delimiter if there is one, don’t
+                            // log an empty line.
+                            let len = line.len();
```
ximon18, line 554: I have a few questions here:
1. Is it guaranteed that the `rsync` command error output uses `\n` as a line terminator?
2. Why not use https://doc.rust-lang.org/std/primitive.slice.html#method.ends_with?
3. Why pop the delimiter rather than just passing a slice range excluding the delimiter to `warn!()`?

partim: 1. You do get CRLF on Windows. Maybe I should indeed do this properly and have different code for Windows.
2. `ends_with` uses a slice as its argument and so does some extra length maths we don’t need.
3. I suppose `pop` just decreases the length by one so it sort of does the same thing. But it does have an additional length check, so given that I did write extra code to save time on 2., maybe I should do this here, too.

ximon18 (approved):

Issue #805, opened 2022-10-15
Docker  Image don't work with IPv6


ruben-herold:
hi,

the docker image is not able to run on ipv6. It only listen to ipv4.

ximon18:
Hi @ruben-herold,

Thanks for the issue report.

My apologies if you are already aware of this but if not it is important to note that Docker [doesn't support IPv6 out-of-the-box](https://docs.docker.com/config/daemon/ipv6/).

Typically with Docker containers the listening is restricted to local ports within the container itself and listen ports are "published" to the host so that connections to the host port are forwarded to the container port. In the case of Routinator this would be the HTTP and RTR service ports (by default 3323 and 8323 when using the defaults with the Routinator Docker image).

I am able to run Routinator on an IPv6 capable host and in a container that has an IPv6 address and IPv6 connectivity to the outside world, e.g.: _(in this case my test host already has a correctly configured Docker daemon to enable IPv6 use within containers)_

```
$ docker volume create routinator-tals
routinator-tals

$ docker run --rm \
    -v routinator-tals:/home/routinator/.rpki-cache/tals nlnetlabs/routinator:v0.11.3 \
    init -f --tal ripe
Created local repository directory /home/routinator/.rpki-cache/repository
Installed 1 TALs in /home/routinator/.rpki-cache/tals

$ docker run -d \
    --restart-unless-stopped \
    --name routinator \
    -p 3323:3323 \
    -p 9556:9556 \
    -v routinator-tals:/home/routinator/.rpki-cache/tals \
    nlnetlabs/routinator:v0.11.3

$ docker exec -it routinator ip -6 addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 state UNKNOWN qlen 1000
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
172: eth0@if173: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 state UP 
    inet6 2a02:a467:c141::242:ac11:2/80 scope global flags 02 
       valid_lft forever preferred_lft forever
    inet6 fe80::42:acff:fe11:2/64 scope link 
       valid_lft forever preferred_lft forever

$ docker exec -it routinator ping -6 google.com
PING google.com (2a00:1450:400e:810::200e): 56 data bytes
64 bytes from 2a00:1450:400e:810::200e: seq=0 ttl=59 time=4.978 ms
[...]

PR #934, opened 2024-02-05
Dockerfile: listen on both IPv4 and IPv6

rissson:
Follow up for https://github.com/NLnetLabs/routinator/issues/805.

Docker indeed doesn't support IPv6 properly, but a lot of other container runtimes do. This change makes routinator running in a container listen on both IPv4 (if available) and IPv6 (if available) by default.

Should I remove the documentation that was added in 603e44dc72a2b5542684969acdac1a51f0b13478 with this change?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -175,4 +175,4 @@ EXPOSE 9556/tcp
 # foreground without the Docker argument "--init" (which is actually another
 # way of activating Tini, but cannot be enabled from inside the Docker image).
 ENTRYPOINT ["/sbin/tini", "--", "routinator"]
-CMD ["server", "--rtr", "0.0.0.0:3323", "--http", "0.0.0.0:8323", "--http", "0.0.0.0:9556"]
+CMD ["server", "--rtr", "[::]:3323", "--http", "[::]:8323", "--http", "[::]:9556"]


ximon18:
Hi @rissson,

Thanks for the contribution, very appreciated! :clap: 

It seems on the surface like a harmless improvement. Howevr, I'm not (yet) convinced that this is a change that we want to make.

The Docker container runtime is the intended target for the `Dockerfile` and the images we create and I am concerned that it could be misleading to make it appear to work with IPv6 out of the box given that IPv6 is not supported by Docker by default.

Also, we have no experience with other container runtimes and prefer where possible not to maintain solutions which we do not test and/or are not familiar with ourselves.

One possible way forward could be to extend our automated testing in some way to verify that this does indeed (a) not break with stock Docker and (b) works as intended under some other container runtime, but that would also potentially imply that we support such other runtimes, and at present I don't think we have the resources to test this ourselves or to extend our test / packaging processes in such a way.

@partim? @AlexanderBand? What are your thoughts on this? Perhaps I am being too strict and we should just say LGTM?

AlexanderBand:
If Docker doesn't support IPv6 by default, but it could possibly work then I'd rather suggest in the documentation that `"[::]:3323"` may work on certain (or most?) systems, rather than making it the default.

rissson:
> The Docker container runtime is the intended target for the `Dockerfile` and the images we create and I am concerned that it could be misleading to make it appear to work with IPv6 out of the box given that IPv6 is not supported by Docker by default.
>
> Also, we have no experience with other container runtimes and prefer where possible not to maintain solutions which we do not test and/or are not familiar with ourselves.

That's fair enough, but one can still use IPv6 with Docker. Either with Docker directly or by using it as a CRI in k8s. I myself have been running Docker with IPv6 turned on, by setting `fixed-cidr-v6` and `ipv6` in `/etc/docker/daemon.json` and running the container with `--sysctl net.ipv6.conf.all.disable_ipv6=0`, and am now in the process of moving stuff over to a dual stack k8s cluster running containerd instead of Docker.

I was banging my head around figuring out why I couldn't reach routinator over IPv6, but could over IPv4, and checking that routinator was listening over IPv6 was the last thing I checked because I assumed that it would. I think most things I run listen on both v4 and v6 if available by default.

On the topic of distributing OCI images, I think it's safe to assume that people will not only run it with the Docker runtime, as they may choose another one because of Docker's (many) flaws.

> One possible way forward could be to extend our automated testing in some way to verify that this does indeed (a) not break with stock Docker and (b) works as intended under some other container runtime, but that would also potentially imply that we support such other runtimes, and at present I don't think we have the resources to test this ourselves or to extend our test / packaging processes in such a way.

I have actually tried that change out with Docker and `--sysctl net.ipv6.conf.all.disable_ipv6=1` to make sure that it still works in an IPv4-only environment, and it still does indeed.

This change shouldn't break existing installations, and allows for a step less for users. Also, if in the future more arguments are added to the default `CMD`, they may be missed by users upgrading that are overriding those to enable IPv6 support.

> If Docker doesn't support IPv6 by default—but it could possibly work—then I'd rather suggest in the documentation that `"[::]:3323"` may work on certain (or most?) systems, rather than making it the default.

Just to make sure we're on the same page, `[::]:3323` means listen on `0.0.0.0:3323` if IPv4 is enabled on the system and `[::]:3323` if IPv6 is enabled on the system. Or rather than system, I should say in the network namespace routinator is running in.

ximon18:
Apologies for the delayed response, it's a bit busy here at the moment. I'm not comfortable making this change at the moment, I'd want to do some testing and I don't have time unfortunately right now.

> Just to make sure we're on the same page, `[::]:3323` means listen on `0.0.0.0:3323` if IPv4 is enabled on the system and `[::]:3323` if IPv6 is enabled on the system. Or rather than system, I should say in the network namespace routinator is running in.

@AlexanderBand: I think this was directed at you?

PR #945, opened 2024-03-19
Consider manifest cert for chain validity and note when points go stale.

### LINKED ISSUE #931: jsonext does not seem to include EE/crl/manifest expiry time
When looking at the `chainValidity` in the `jsonext` output it seems like manifest/crl expiry is not considered. It might be nice to include this information in the result so a user can see how long an object currently really is valid.

Very low priority.

```json
// curl https://rpki-validator.ripe.net/jsonext 2>/dev/null | grep rpki.ripe.net | head -n 10 | tail -n 1 | sed -e's/,$//' | jq '.'
{
  "asn": "AS12479",
  "prefix": "1.178.248.0/21",
  "maxLength": 21,
  "source": [
    {
      "type": "roa",
      "uri": "rsync://rpki.ripe.net/repository/DEFAULT/a7/1a830a-f061-4cdc-bafb-a2fe9f015d71/1/lTI6zlhbGxM_MiAwrBp26vfY0BM.roa",
      "tal": "ripe",
      "validity": {
        "notBefore": "2024-01-23T10:07:11Z",
        "notAfter": "2025-07-01T00:00:00Z"
      },
      "chainValidity": {
        "notBefore": "2024-01-23T14:32:31Z",
        "notAfter": "2024-07-01T00:00:00Z"
      }
    }
  ]
}
```

partim:
This PR changes the code that calculates the validity of the full validation chain to also consider the validity of the EE certificates of the publication points along the chain. It also adds a new field in the jsonext output format that provides the time when any of the publication points along the validation chain become stale.

Fixes #931.

### DIFF
--- a/src/payload/validation.rs
+++ b/src/payload/validation.rs
@@ -133,7 +133,8 @@ impl<'a> ProcessRun for &'a ValidationReport {
             PubPointProcessor {
                 report: self,
                 pub_point: PubPoint::new_ta(cert, tal_index),
-                validity: cert.combined_validity(),
+                validity: cert.cert().validity(),
+                point_stale: cert.cert().validity().not_after(),
             }
         ))
     }
@@ -155,8 +156,11 @@ pub struct PubPointProcessor<'a> {
     /// The data being collected.
     pub_point: PubPoint,
 
-    /// The (combined) validity of the CA certificate.
+    /// The (combined) validity of the publication point.
     validity: Validity,
+
+    /// When will this publication point become stale.
+    point_stale: Time,
 }
 
 impl<'a> ProcessPubPoint for PubPointProcessor<'a> {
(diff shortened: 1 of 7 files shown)


AlexanderBand (approved):

src/payload/validation.rs
```diff
@@ -155,19 +156,24 @@ pub struct PubPointProcessor<'a> {
...
 impl<'a> ProcessPubPoint for PubPointProcessor<'a> {
     fn repository_index(&mut self, repository_index: usize) {
         self.pub_point.repository_index = Some(repository_index)
     }
 
-    fn update_refresh(&mut self, not_after: Time) {
+    fn point_validity(&mut self, manifest: Validity, stale: Time) {
         self.pub_point.refresh = cmp::min(
-            self.pub_point.refresh, not_after
+            self.pub_point.refresh, manifest.not_after()
```
DRiKE, line 173: Shouldn't this also consider `stale`?

DRiKE (approved):

PR #946, opened 2024-03-20
Store manifest number and check for regressions.

### LINKED ISSUE #913: Prefer highest numbered, valid, complete manifest
After some discussion today I think that if multiple copies exist (current content, cache) the most recent (highest number) manifest should be used.

If the cache, and current repository both contain a manifest that:

  * passes all relevant checks
  * and thus has a complete set of files present
  * and thus is in its validity window
 
 
Routinator should pick the manifest with the highest number, and log if the repo contained an older number than the cache.

This is a trade off: this picks the legit object during a replay attack. However it also sticks to the wrong object when a CA traveled back in time after a operational issue. This is not a security issue, but this implementation is closer to rfc9268.

partim:
This PR adds a check for manifest number regressions when validating a collected publication point. It stores the manifest number for each manifest and checks against it when collecting a new manifest. If the latter’s number has not increased, it falls back to the stored manifest. This behaviour is mandated by RFC 9286.

The PR changes the data stored for manifests and thus updates the `StoredManifest` version to 1. In order to avoid an endless stream of error messages after an upgrade, it downgrades the logged message when encountering an malformed `StoredManifest` to DEBUG.

Fixes #913.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,9 +34,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.12"
+version = "0.6.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96b09b5178381e0874812a9b157f7fe84982617e48f71f4e3235482775e5b540"
+checksum = "d96bd03f33fe50a863e394ee9718a706f988b9079b20c3784fb726e7678b62fb"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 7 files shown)


DRiKE (approved):

DRiKE:
Just wondering: any the longer term plan for the loglevel of that message? I imagine you'd actually want it to be INFO again at some point.

partim:
We should probably collate these error messages into just one which then can be _warn._ There’s a vague plan to rework logging, so it should be done then.

PR #951, opened 2024-04-09
Ensure that RRDP delta hashes don’t changes between updates.

partim:
This PR ensures that the hash of an RRDP delta with a given serial doesn’t change between updates. It stores the list of delta serials and hashes with the RRDP repository state in its archive and checks that hashes for serial numbers present both in the repository state and a new notification are equal. Otherwise falls back to a snapshot update.

This PR implements the [draft-ietf-sidrops-rrdp-desynchronization-00](https://datatracker.ietf.org/doc/draft-ietf-sidrops-rrdp-desynchronization/). The draft suggests to limit the number of deltas stored. We are not yet doing that. Instead this should be part of limiting the number of deltas taken out of the notification file when parsing in a follow up PR in rpki-rs.

This PR changes the format of the repository state and thus increases its version to 1. Strictly speaking, we never released version 0, but it’s been in _main_ from quite some time, so an increase feels prudent.

### DIFF
--- a/src/utils/binio.rs
+++ b/src/utils/binio.rs
@@ -4,7 +4,8 @@
 //! how to serialize themselves. The module implements the traits for all the
 //! types we need.
 
-use std::{error, fmt, io, slice};
+use std::{error, fmt, hash, io, slice};
+use std::collections::HashMap;
 use bytes::Bytes;
 use rpki::{rrdp, uri};
 use rpki::repository::x509::Serial;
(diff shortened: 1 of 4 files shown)


src/utils/binio.rs
```diff
@@ -329,6 +330,44 @@ impl<R: io::Read> Parse<R> for Serial {
...
 
 
+//------------ HashMap<K, V> -------------------------------------------------
+//
+// Encoded as the number of items as a u64 followed by pairs of key and value.
+
+impl<K: Compose<W>, V: Compose<W>, W: io::Write> Compose<W> for HashMap<K, V> {
+    fn compose(&self, target: &mut W) -> Result<(), io::Error> {
+        u64::try_from(self.len())
+        .map_err(|_| ParseError::format("excessively large vec"))?
```
DRiKE, line 340:
```suggestion
        .map_err(|_| ParseError::format("excessively large map"))?
```

src/utils/binio.rs
```diff
@@ -329,6 +330,44 @@ impl<R: io::Read> Parse<R> for Serial {
...
+
+impl<K, V, R> Parse<R> for HashMap<K, V>
+where
+    K: Parse<R> + Eq + hash::Hash,
+    V: Parse<R>,
+    R: io::Read
+{
+    fn parse(source: &mut R) -> Result<Self, ParseError> {
+        let len = usize::try_from(u64::parse(source)?).map_err(|_| {
+            ParseError::format("too many items in vec")
```
DRiKE, line 358:
```suggestion
            ParseError::format("too many items in map")
```

DRiKE (approved):

PR #953, opened 2024-04-10
Enforce same origin for all resources of an RRDP server.

partim:
This PR enforces that all resources fetched for an RRDP server have the same origin as the URI provided in the CA certificate. It checks this for all URIs provided in the server’s notification file and restricts redirects to URIs with the same origin.

### DIFF
--- a/src/collector/rrdp/http.rs
+++ b/src/collector/rrdp/http.rs
@@ -5,7 +5,7 @@ use std::time::Duration;
 use bytes::Bytes;
 use chrono::{DateTime, Utc};
 use log::{error, warn};
-use reqwest::header;
+use reqwest::{header, redirect};
 use reqwest::{Certificate, Proxy, StatusCode};
 use reqwest::blocking::{Client, ClientBuilder, RequestBuilder, Response};
 use rpki::uri;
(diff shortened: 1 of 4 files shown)


src/collector/rrdp/http.rs
```diff
@@ -256,6 +259,29 @@ impl HttpClient {
...
+    ///
+    /// We allow up to 10 redirects (reqwest’s default policy) but only if
+    /// the origin stays the same.
+    fn redirect_policy(attempt: redirect::Attempt) -> redirect::Action {
+        if attempt.previous().len() > 9 {
+            return attempt.stop();
+        }
+        let orig = match attempt.previous().first() {
+            Some(url) => url,
+            None => return attempt.follow() // Shouldn’t happen?
```
DRiKE, line 273: Is there no more conservative alternative instead of `.follow()`? If this is really unexpected, perhaps a warning should be logged. Perhaps combined with a `return attempt.stop()`?
A warning might actually be quite useful for (repo) operators to troubleshoot presumably unexpected behavior.

DRiKE (approved):

src/collector/rrdp/http.rs
DRiKE: If the alternative is to (silently) stop processing this RRDP repo, maybe this is better indeed.

PR #961, opened 2024-06-06
Limit the maximum number of deltas in an RRDP notification.

partim:
This PR limits the number of entries in the notification file of an RRDP repository that Routinator is willing to consider. This can be set through the new `rrdp-max-delta-list-len` configuration value which defaults to a hopefully safe 500.

If this number is exceeded, the delta list will be considered to be empty and ignore, resulting in snapshots to be used.

### DIFF
--- a/doc/manual/source/manual-page.rst
+++ b/doc/manual/source/manual-page.rst
@@ -279,6 +279,13 @@ The available options are:
       larger than the value provided by this option, the snapshot is used
       instead. If the option is missing, the default of 100 is used.
 
+.. option:: --rrdp-max-delta-list-len=len
+ 
+      If the number of deltas included in the notification file of an RRDP
+      repository is larger than the value provied, the delta list is
+      considered empty instead and the snapshot is used. If the option is
+      missing, the default of 500 is used.
+
 .. option:: --rrdp-timeout=seconds
 
       Sets the timeout in seconds for any RRDP-related network operation,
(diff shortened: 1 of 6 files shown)


DRiKE (approved):

doc/manual/source/manual-page.rst
```diff
@@ -279,6 +279,13 @@ The available options are:
       larger than the value provided by this option, the snapshot is used
       instead. If the option is missing, the default of 100 is used.
 
+.. option:: --rrdp-max-delta-list-len=len
+ 
+      If the number of deltas included in the notification file of an RRDP
+      repository is larger than the value provided, the delta list is
+      considered empty instead and the snapshot is used. If the option is
```
AlexanderBand, line 286:
```suggestion
      considered empty and the snapshot is used instead. If the option is
```

doc/manual/source/manual-page.rst
```diff
@@ -1119,6 +1126,12 @@ All values can be overridden via the command line options.
             necessary to update an RRDP repository before using the snapshot
             instead. If the value is missing, the default of 100 is used.
 
+      rrdp-max-delta-list-len
+            An integer value that specifies the maximum number of deltas
+            listed the notification file of an RRDP repository before the
+            list is considered empty instead and the snapshot is user.
```
AlexanderBand, line 1132:
```suggestion
            list is considered empty and the snapshot is used instead.
```

AlexanderBand (changes requested):

AlexanderBand (approved):

PR #962, opened 2024-06-07
Add --no-motd to and remove -l from rsync options.

partim:
This PR adds the `--no-motd` option to the default extra rsync options and removes `-l` from the base rsync options.

This means we now have `-rtO --delete` as the base options and `-z`, `--contimeout=10`, `--max-size`, and `--no-motd` as the default extra options that can be overwritten.

I think keeping `--delete` is fine – an attacker can always replace a file rather than deleting it and still break the publication point in question. We cannot limit the file patterns accepted as that would prevent adding new file types until all relying party installations are updated – RFC 9286 specifically says that missing fails lead to a failed fetch.

### DIFF
--- a/doc/manual/source/manual-page.rst
+++ b/doc/manual/source/manual-page.rst
@@ -1089,8 +1089,10 @@ All values can be overridden via the command line options.
             The options ``-rtzO --delete`` are always passed to the command.
             The options listed in the option are added to it.
 
-            If the option is not provided, will add ``--contimeout=10`` if
-            that is supported by the rsync command and ``--no-motd`` always.
+            If the option is not provided, will add ``--contimeout=10``, if
+            it is supported by the rsync command, `--max-size` if the
+            ``max-object-size`` option has not been set to 0, and
+            ``--no-motd`` always.
 
       rsync-timeout
             An integer value specifying the number seconds an rsync command
--- a/doc/manual/source/manual-page.rst
+++ b/doc/manual/source/manual-page.rst
@@ -1083,14 +1083,14 @@ All values can be overridden via the command line options.
             default is simply *rsync*.
 
       rsync-args
-            A list of strings containing the arguments to be passed to the
-            rsync command. Each string is an argument of its own.
+            A list of strings containing additional arguments to be passed
+            to the rsync command. Each string is an argument of its own.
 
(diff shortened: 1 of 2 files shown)


DRiKE (approved):

Koenvh1 (approved):
Looks good to me. Not sure how much the -0 adds, and adding that does break things for people that use openrsync and link that as rsync (does anyone do that?).

```
koen@beta:~/Downloads/openrsync$ ./openrsync -rtz0 rsync://rpki.ripe.net/ta /tmp/ripenccrsync/
openrsync: -z not supported yet
./openrsync: invalid option -- '0'
usage: openrsync [-aDglnoprtvx] [-e program] [--address=sourceaddr]
	[--compare-dest=dir] [--del] [--exclude] [--exclude-from=file]
	[--include] [--include-from=file] [--no-motd] [--numeric-ids]
	[--port=portnumber] [--rsync-path=program] [--timeout=seconds]
	[--version] source ... directory
```

partim:
It’s a capital letter O (yeah, using that is a bit of a bad idea), so the shortcut for `--omit-dir-times`. It’s probably not a big deal to add it or not.

partim:
Maybe we should move the `-z` into the default options that cab be overwritten?  That would make it possible to use openrsync without getting errors all the time.

AlexanderBand (approved):

doc/manual/source/manual-page.rst
```diff
@@ -1083,14 +1083,16 @@ All values can be overridden via the command line options.
...
-            ten seconds. If your rsync is too old to support this option, no
-            arguments are used.
+            A list of strings containing additional arguments to be passed
+            to the rsync command. Each string is an argument of its own.
+
+            The options ``-rtzO --delete`` are always passed to the command.
+            The options listed in the option are added to it.
+
+            If the option is not provided, will add ``--contimeout=10``, if
+            it is supported by the rsync command, `--max-size` if the
```
AlexanderBand, line 1093:
```suggestion
            it is supported by the rsync command, ``--max-size`` if the
```

AlexanderBand (approved):

doc/manual/source/manual-page.rst
partim: Fixed.

PR #972, opened 2024-07-02
Remove EOL systems

AlexanderBand:


AlexanderBand:
Also, do we build packages for Ubuntu 24.04 'Noble Numbat' (LTS) yet? It was released on 25 Apr 2024.

ximon18:
No, that hasn't been added yet. (and would also need adding to the packaging backend, not just in Routinator GH).

PR #979, opened 2024-11-16
Fedora has switched to SPDX licenses since 40

pemensik:
All new packages should have lincese in SPDX format on Fedora. There is no need for separate license in legacy format, that would be considered wrong.

On Fedora package, the license field should include all binary dependencies built into binary packages. Not just the project itself. It is not strictly neccessary until you try to have your generated rpm in official repos. But you might be aware of that. That is the cost of bundling all dependencies in form of static libraries.

Details on:
- https://docs.fedoraproject.org/en-US/packaging-guidelines/LicensingGuidelines/#_license_field
- https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/#_license_tags

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,10 +93,6 @@ systemd-units = { unit-name = "routinator", unit-scripts = "pkg/common", enable 
 [package.metadata.deb.variants.minimal]
 
 [package.metadata.generate-rpm]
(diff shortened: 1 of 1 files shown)


partim:
Thank you for the PR and apologies for the delay.

I have now confirmed that the change doesn’t break our own Ploutos-based builds, so it can be merged.

PR #982, opened 2024-12-06
Improve archive file sizes.

partim:
~~This PR attempts to improve the growth of (RRDP) archive files. It changes the strategy for updating objects that have grown in size to by trying to re-use empty space rather than just appending them.~~

This PR improves the growth of RRDP archive files. It tries to re-use existing empty space when updating objects. It also pads all objects to a multiple of 256 bytes to make it more likely that empty space can be reused.

The PR also adds a new command `archive-stats` that prints statistics for an archive file.

Finally, the PR raises the Minimum Supported Rust Version to 1.73 to be able to use `u64::next_multiple_of`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        rust: [1.70.0, stable, beta]
+        rust: [1.73.0, stable, beta]
     steps:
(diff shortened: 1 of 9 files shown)


partim:
The PR has now been updated to also pad all objects to a multiple of 256 bytes. This seriously helps with reuse of empty space at the price of every object being a bit larger. It turned out in the long run that this helped quite a bit with curbing file growth.

Koenvh1 (approved):

PR #989, opened 2025-01-02
Introduce a limit for ASPA provider ASNs.

partim:
This PR introduces a new configuration option `aspa-provider-limit` that limits the number of provider ASNs allowed in an ASPA object. If the number is exceeded, all ASPA assertions for the customer ASN are removed from the dataset to avoid false rejections. The default value is 10,000.

### DIFF
--- a/src/payload/validation.rs
+++ b/src/payload/validation.rs
@@ -15,7 +15,7 @@
 
 use std::cmp;
 use std::collections::hash_map;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use crossbeam_queue::SegQueue;
 use log::{info, warn};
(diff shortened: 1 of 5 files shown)


src/payload/validation.rs
```diff
@@ -271,14 +275,22 @@ impl ProcessPubPoint for PubPointProcessor<'_> {
...
         cert: ResourceCert,
         aspa: AsProviderAttestation
     ) -> Result<(), Failed> {
         if !self.report.enable_aspa {
             return Ok(())
         }
         self.pub_point.update_refresh(cert.validity().not_after());
+        if aspa.provider_as_set().len() > self.report.aspa_provider_limit {
+            warn!(
+                "{}: too many provider ASNs. Skipping ASPA for {}.",
```
Koenvh1, line 288: Given that you have access to both `self.report.aspa_provider_limit` and `aspa.provider_as_set().len()` here anyway, maybe it's nice to include it in the warning message?

Koenvh1 (approved):

src/payload/validation.rs
partim: Fair point. I’ll fix it.

PR #1010, opened 2025-02-06
Remove aspa cargo feature from docs, ref #990

mxsasha:
ASPA is always compiled in from 0.14.1 through #990, but the docs were not updated.


partim:
Thank you for the PR!

Issue #1025, opened 2025-04-12
Inquiry Regarding Time Intervals Between Consecutive Updates in Routinator


Meadows1254465703:
I am writing to share some observations regarding the time intervals between consecutive validation runs in Routinator, based on our recent analysis of its log outputs. 

To investigate the update intervals, we calculated the time differences between the start times of consecutive validation runs by analyzing log entries marked with "Starting a validation run." This allowed us to derive the distribution of these intervals.

We conducted this analysis using Routinator v0.14.2 with default parameters (including the documented refresh value of 10 minutes, which should delay the next update until 10 minutes after the previous run completes). Notably, the observed intervals exhibited a bimodal distribution across two independent machines:

• Machine 1: 

![Image](https://github.com/user-attachments/assets/853f666f-5fd8-43d3-8e60-4321ca584ba5)

![Image](https://github.com/user-attachments/assets/7c2cfffd-c5ce-4016-9512-5dd69b0e2b4b)

• Machine 2:

![Image](https://github.com/user-attachments/assets/1762a57b-f630-48ad-8c05-734e3d202825)

![Image](https://github.com/user-attachments/assets/6a7636d9-0458-4bc9-9b80-f33e55764f35)

While the second peak in the distribution aligns with the expected 10-minute refresh interval described in the documentation, the first peak suggests unexpected behavior. Further examination of the logs revealed frequent occurrences of the entry:
[INFO] Next validation run scheduled in 0 seconds
This implies that certain validation runs initiated immediately after the previous run concluded, deviating from the intended 10-minute delay.

Could you kindly help us understand the following:
1.What scenarios might trigger a validation run to restart with a 0-second delay despite the default refresh setting?
2.Are there additional configurations or external factors that could influence this behavior?

We sincerely appreciate your expertise and time in addressing these questions. Please let us know if further details or logs would assist your investigation.

partim:
Hi! Thank you for doing this research and reaching out! We have had reports of this before but somehow haven’t gotten around to properly figuring out how and why this happens. As you can imagine, this is quite tricky to get to the ground of.

One question that came up discussing this here just now was whether these 0-second waits happen in sequence or if typically the next wait is normal. Would you perhaps be willing to provide us with the raw data you did your calculations on? That would perhaps allow us to discover other patterns as well. If you don’t want to publish the data, you can also send it to us privately at rpki-team@nlnetlabs.nl.

Koenvh1:
[...]

PR #1027, opened 2025-04-25
Change Routinator refresh behaviour

Koenvh1:
This PR adds support for a min-refresh option in the configuration, and changes the default behaviour relating to #1025

Currently Routinator will look at the earliest expiry time of any non-expired object in the RPKI, and set the refresh time to that time. The validation of all objects in the RPKI takes some time. In this time, an object that was already verified as valid might have expired before the run completes. This would cause Routinator to immediately run again after completion.

This PR changes that behaviour by introducing a minimum duration to wait, and if not set to wait the `refresh` time (by default 10 minutes).

### DIFF
--- a/doc/manual/source/manual-page.rst
+++ b/doc/manual/source/manual-page.rst
@@ -856,6 +856,21 @@ These can be requested by providing different commands on the command line.
               The amount of seconds the server should wait after having
               finished updating and validating the local repository before
               starting to update again. The next update will be earlier if
+              objects in the repository expire earlier and min-refresh is set. 
+              The default value is 600 seconds.
+
+       .. option:: --min-refresh=seconds
+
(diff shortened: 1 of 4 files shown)


doc/manual/source/manual-page.rst
```diff
@@ -1331,8 +1331,13 @@ All values can be overridden via the command line options.
       refresh
             An integer value specifying the number of seconds Routinator
             should wait between consecutive validation runs in server mode.
-            The next validation run will happen earlier, if objects expire
-            earlier. The default is 600 seconds.
+            The next validation run will happen earlier if objects expire
+            earlier and min-refresh is set. The default is 600 seconds.
+
+      min-refresh
```
partim, line 1337: The command line version of the setting also needs to be documented.

partim (approved):

PR #1036, opened 2025-06-26
Add support for ARMv7 / ARM64

Koenvh1:
Based on a manual test of https://github.com/NLnetLabs/routinator/actions/runs/15905341492 on a Hetzner Debian 12 ARM64 machine it seems to work fine.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -118,7 +118,7 @@ fn main() {
 
     let mut assets = Assets::default();
     if let Err(err) = assets.load_dir(UI_DIR.into()) {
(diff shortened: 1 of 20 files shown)


partim (approved):
Do I see it correctly that these new builds are excluded from testing?

Koenvh1:
@partim They are not tested, no, because Incus does not support ARM64 images on an x64 host: https://discuss.linuxcontainers.org/t/arm64-container-vm-on-x86-64-host/19583/5

PR #1040, opened 2025-06-30
Update dependencies.

partim:
This PR updates the dependencies. The only interesting thing is the replacement of `verify_slices_are_equal` with a simple comparison which should be fine when comparing the SHA256 hashes of RRDP snapshots and deltas.


partim:
For some reason, doing a `cargo update` triggered more Clippy complaints on the same compiler version in otherwise unchanged code. So these Clippy fixes are here, too.

Koenvh1 (approved):

PR #1041, opened 2025-07-01
Randomise processing of file list

Koenvh1:
This adds randomisation of the manifest processing as discussed in https://mailarchive.ietf.org/arch/msg/sidrops/gaX6rFkiwBfKKeE6PXAu2Pz8BIA/

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -745,13 +745,12 @@ impl<'a, P: ProcessRun> PubPoint<'a, P> {
         // signal that the publication point was processed successfully but
         // shouldn’t be considered further.
         let mut ca_tasks = Vec::new();
-        let items = collected.content.iter();
 
         // Randomise the manifest to ensure a non-predictable ordering of the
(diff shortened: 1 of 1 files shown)


src/engine.rs
```diff
@@ -744,7 +745,16 @@ impl<'a, P: ProcessRun> PubPoint<'a, P> {
...
         // shouldn’t be considered further.
         let mut ca_tasks = Vec::new();
-        let mut items = collected.content.iter();
+        let items = collected.content.iter();
+
+        // Randomise the manifest to ensure a non-predictable ordering of the
+        // child CAs. This is a preventative measure to stop CA children from
+        // trying to influence the processing of the other children. It also
+        // adds randomness to visiting the repositories, reducing peak load.
+        let mut items_random = Vec::from_iter(items);
```
partim, line 754: You can shorten by using `Iterator::collect`, i.e., something like (not checked):

```
let mut items_random: Vec<_> = collected.content.iter().collect();
```

PR #1043, opened 2025-07-11
Use Stream Buffers to Read/Write validation input/outputs

kawaemon:
This pull request improves validation input/output file operations' performance, by using buffers.
JSON readers/writers are tend to do small file system operations very frequently, so buffering is essential.

Simple benchmark shows 5% win in total (including startup ROA validations):

image

### DIFF
--- a/src/operation.rs
+++ b/src/operation.rs
@@ -891,8 +891,8 @@ impl Validate {
         let result = requests.validity(&snapshot);
         match self.output.as_ref() {
             Some(path) => {
-                let mut file = match fs::File::create(path) {
-                    Ok(file) => file,
+                let mut stream = match fs::File::create(path) {
+                    Ok(file) => io::BufWriter::new(file),
                     Err(err) => {
                         error!(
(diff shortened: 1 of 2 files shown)


src/operation.rs
```diff
@@ -805,7 +805,7 @@ impl Validate {
                 Ok(validity::RequestList::single(prefix, asn))
             }
             ValidateWhat::File(ref path) => {
-                let mut file = match fs::File::open(path) {
+                let file = match fs::File::open(path) {
                     Ok(file) => file,
```
partim, line 809: Since we don’t need `file` to be a `File` further down, you can just wrap it in a `BufReader` right here in the Ok branch.

(Same also further down.)

partim (changes requested):
Thank you for your PR! Looks good with some minor nits.

src/operation.rs
kawaemon: alright, now fixed!

partim (approved):
Looks good now. Again, thank you for your PR!

PR #1048, opened 2025-08-01
Check whether Routinator works on Windows 2025

Koenvh1:
Relates to https://github.com/actions/runner-images/issues/12677


Koenvh1:
Seems to work fine - nothing we need to do.

PR #1049, opened 2025-08-07
Introduce `vrps --update-after`.

### LINKED ISSUE #88: Feature-request: Allow the -n switch to have a argument to update if last update is more than X seconds ago
I have a script that runs routinator and then creates config based on the json output. It currently updates everything, including rsync. I can check the age of the current data in the script, but it might be useful for other users as well if the '-n' switch has an optional argument. If the age of the data is mode than 'argument' seconds old, update it, otherwise, use the current data.

partim:
This PR introduces a new command line option to the `vrps` command, `--update-after=minutes` that disables updating the local repository cache if the last update was less than the given number of minutes ago.

Fixes #88. (Better late then never.)


partim:
I figured minutes is likely what a user wants and multiplying by 60 can get annoying?

PR #1057, opened 2025-09-02
Introduce an initial run without collectors.

### LINKED ISSUE #728: In server mode, if stored data is not too old, start without refreshing.
(Re-) starting Routinator by now takes ages. To get around this, check if it is possible to do an initial validation run on stored data only if that is not older than the refresh time.

partim:
This PR introduces the option for what we call an “initial run” which runs without the collectors enabled and aborts as soon as a publication point is requested that hasn’t been requested before. This is intended to allow the server to quickly create a dataset after a restart.

Fixes #728.

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -816,16 +816,29 @@ impl StoredPoint {
             // We never succeeded. Update the status and return.
             header.update_status = UpdateStatus::LastAttempt(Time::now());
 
+            drop(file);
+            let mut file = File::create(&path).map_err(|err| {
+                error!(
+                    "Failed to update stored publication point at {}: \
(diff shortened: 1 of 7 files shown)


src/store.rs
```diff
@@ -132,7 +157,7 @@ impl Store {
     ///
     /// The function is called implicitly by [`new`][Self::new].
     //  (Or, well, not really, but they both only call `create_base_dir`, so
-    //   from a user persepective it does.)
+    //   from a user perspeective it does.)
```
Koenvh1, line 160: I believe this changes one typo into another typo :-)

Koenvh1 (approved):

PR #1061, opened 2025-09-12
Add rrdp-read-timeout and revise timeout defaults.

partim:
This PR adds a new `rrdp-read-timeout` config parameter which sets the timeout for all network operations related to RRDP. The default for this timeout is 10 seconds. As this is also used for the initial connect, this will make broken servers timeout much more quickly.

In addition, the PR changes the default for the `rrdp-timeout` value – which concerns the full exchange – to 600 seconds as we have seen some repositories time out the snapshot with the previous default of 300 seconds. We think it beneficiary to give them more time to finish the snapshot update rather than falling back to rsync and then trying again and failing again after five minutes for every validation run.

One question: Do we think 10 seconds is a good default for the `rrdp-read-timeout`? The default used by _reqwest_ is 30 seconds, but this seems to be really quite generous?

### DIFF
--- a/doc/manual/source/manual-page.rst
+++ b/doc/manual/source/manual-page.rst
@@ -290,15 +290,21 @@ The available options are:
 
 .. option:: --rrdp-timeout=seconds
 
-      Sets the timeout in seconds for any RRDP-related network operation,
-      i.e., connects, reads, and writes. If this option is omitted, the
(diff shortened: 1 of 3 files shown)


AlexanderBand (changes requested):
Some typos

PR #1076, opened 2025-10-28
Improve performance of batched route validation

kawaemon:
This PR improves time complexity of route validation from O(n^2) to O(n log(n)) by using [prefix-trie](https://docs.rs/prefix-trie/latest/prefix_trie/).

As shown in the benchmark section, building tree takes some time, and linear search is faster when processing fewer than 100 routes. So in this PR it uses tree if batch size is large enough.

### Benchmark

Code is available [here](https://github.com/kawaemon/routinator/tree/bench).

| Routes   | Old Avg (ms) | New Avg (ms) | Speedup (x) |
|---------:|--------------:|-------------:|------------:|
| 1        | 1.764         | 143.822      | 0.012       |
| 10       | 17.272        | 116.801      | 0.148       |
| 100      | 172.341       | 116.962      | 1.473       |
| 1000     | 1719.030      | 117.308      | 14.654      |
| 10000    | 17263.463     | 119.528      | 144.431     |

The tree uses approximately 230MB of memory. I'd like to know whether this is acceptable.

```
> # old
> /usr/bin/time -f 'took %E, used %M KB' ./target/release/routinator validate --noupdate -i <(head -n 100000 ./.tmp/fullroute.routinator.txt) -o /dev/null
took 2:59.33, used 434016 KB

> # new
> /usr/bin/time -f 'took %E, used %M KB' ./target/release/routinator validate --noupdate -i <(head -n 100000 ./.tmp/fullroute.routinator.txt) -o /dev/null
took 0:08.83, used 663020 KB
```

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,8 @@ Bug fixes
 
 Other changes
 
+* Improved performance of batched route validation. ([#1076])
+
 
 ## 0.15.1 ‘Ain’t No Country Club Either’
 


kawaemon:
> The plan would be to actually keep the ROA data in a prefix trie of some form,
	
Indeed that's the cleanest way.

I'll give it a shot and see how it goes... Thank you for explanation!

kawaemon:
Looks like we need several breaking changes including rpki-rs crate. So I will close this for now, and I will post new set of PRs if found enough working time.

PR #1077, opened 2025-10-31
Add RSC support

Koenvh1:
This PR adds support for Resource Signed Checklists ([RFC 9323](https://datatracker.ietf.org/doc/rfc9323/)) in Routinator.

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   package:
-    uses: NLnetLabs/ploutos/.github/workflows/pkg-rust.yml@v8
+    uses: NLnetLabs/ploutos/.github/workflows/pkg-rust.yml@v9
     secrets:
       DOCKER_HUB_ID: ${{ secrets.DOCKER_HUB_ID }}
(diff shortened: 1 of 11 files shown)


src/operation.rs
```diff
@@ -1142,6 +1149,164 @@ impl ValidateDocument {
...
+                    } else {
+                        continue 'outer;
+                    }
+                }
+            }
+
+            // At this point none of the entries in the check list matched the
+            // document we supplied.
+
+            error!("Failed to match document to valid entry in the check list '{}'.",
```
partim: Perhaps collect all the failed document file names into a vec and print all of them if that vec doesn’t end up being empty.

src/rsc.rs
```diff
@@ -0,0 +1,192 @@
...
+
+impl ValidationReport {
+    fn supply_tal(&self, tal: &Tal) -> Result<bool, ValidationError> {
+        if self.complete.load(Ordering::Relaxed) {
+            return Ok(true);
+        }
+
+        let cert = self.rsc.signed().cert();
+
+        if cert.is_self_signed() {
```
partim: This would suggest that the EE cert of the RSC is the TA certificate. I don’t think that can happen?

Issue #1082, opened 2026-03-23
Improve logging


Koenvh1:
e.g. https://community.nlnetlabs.nl/t/routinator-with-lab-tal-fails-to-snyc-via-rrdp/84, the `client error (Connect)` should probably be better

PR #1083, opened 2026-03-23
Try to unroll all error sources

Koenvh1:
Relates to #1082

Makes errors look like:
```
[2026-03-23T15:23:08] [INFO] RRDP https://rrdp.ripe.net/notification.xml: client error (Connect) (tcp connect error (Connection refused (os error 111) (Connection refused (os error 111))))
```

### DIFF
--- a/src/collector/rrdp/base.rs
+++ b/src/collector/rrdp/base.rs
@@ -1,4 +1,3 @@
-use std::error::Error;
 use std::{cmp, fs, io};
 use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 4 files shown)


partim:
I think this can be done without collecting into a string string first: Make a newtype wrapping a reference to the error and `impl Display` for that which can then just write to the formatter.

However and contrary to the above: Maybe we can skip duplicate messages? As in: write the message to a string after all and keep it around for the next iteration. If the next iteration’s string is the same, skip it.

Koenvh1:
I think the latter option is nicer (or less of a mess), which is what I did now.

PR #1090, opened 2026-04-28
Make the TA certificate loading correctly adhere to the `max_object_size` setting even without Content-Length header

Koenvh1:


partim:
For reading RRDP files, you already have [`RrdpDataRead`](https://github.com/NLnetLabs/routinator/blob/main/src/collector/rrdp/update.rs#L493) that does a checked read. We should probably refactor that type into a something that can also be used to read a certificate rather than re-implementing.

partim (approved):

PR #1091, opened 2026-04-28
Add sanity check on passed rsync arguments

Koenvh1:


src/collector/rsync.rs
```diff
@@ -623,6 +623,22 @@ impl RsyncCommand {
         };
         let mut cmd = AsyncCommand::new(&self.command);
         for item in &self.args {
+            if item == "-e" || item == "--rsh" {
+                log.error(format_args!("illegal rsync argument {item}"));
+                return Err(io::Error::other("illegal rsync argument"));
+            }
+            if env::var("RSYNC_RSH").is_ok() {
```
partim: This should probably be outside the for loop?

src/collector/rsync.rs
```diff
@@ -623,6 +623,22 @@ impl RsyncCommand {
...
+                log.error(format_args!(
+                    "illegal rsync environment variable RSYNC_RSH"
+                ));
+                return Err(io::Error::other(
+                    "illegal rsync environment variable"
+                ));
+            }
+            log.warn(format_args!(
```
partim: I’m not sure this needs to be here? Adding arguments is kind of the point of the configuration option?

partim (changes requested):

partim (approved):

partim (changes requested):
It occurs to me now that the checks shouldn’t be in `RsyncCommand::command` but in `RsyncCommand::new`. I.e., this should only be done once.

partim:
Oh, and we should probably update the documentation to say that ”-e” will be rejected.

partim (approved):

PR #1092, opened 2026-04-28
Strengthen rsync module path slicing with potentially malformed module

Koenvh1:


partim:
I’m not keen on the `unwrap_or` which just quietly paves over a bug in unexpected ways. I suspect it would be better to make `Module` contain a `uri::Rsync` or something, but I want to renovate the `rpki::uri` module to not use `Bytes` any more, so maybe we should wait until then?

Koenvh1:
That is probably a neater solution

Issue #1018, opened 2025-03-10
how to add ASPA objets as local exceptions ?


SergeKrier:
pretty much like https://routinator.docs.nlnetlabs.nl/en/stable/local-exceptions.html#local-exceptions

but for ASPA, or any other mean by local config ? 

thanks

AlexanderBand:
As far as I know, the ASPA definition for SLURM is still under dicussion as an IETF Draft:
https://datatracker.ietf.org/doc/html/draft-ietf-sidrops-aspa-slurm

SergeKrier:
true, maybe worth considering early support so can be added to [Appendix A. ](https://datatracker.ietf.org/doc/html/draft-ietf-sidrops-aspa-slurm#appendix-A)[Implementation status](https://datatracker.ietf.org/doc/html/draft-ietf-sidrops-aspa-slurm#name-implementation-status-rfc-e)
just saying...

oskar456:
Hello!

I believe #1021 only added SLURMv2 **output** support, while the OP was asking for support for ASPA exceptions **input**. A quick glance into [slurm.rs](https://github.com/NLnetLabs/routinator/blob/da6a8440309a018ea16adaa545c8d23edf980c6e/src/slurm.rs#L17-L23) shows that it only parses ROA and BGPsec filters and assertions, while it ignores any ASPA filters or assertions:

```rust
pub struct LocalExceptions {
    origin_filters: Vec<PrefixFilter>,
    router_key_filters: Vec<BgpsecFilter>,

    origin_assertions: Vec<(RouteOrigin, Arc<ExceptionInfo>)>,
    router_key_assertions: Vec<(RouterKey, Arc<ExceptionInfo>)>,
}
```

Would it be possible to add ASPA support for local exceptions?

partim:
Er, good point.

Yes. That would certainly be possible and should indeed happen.

PR #1106, opened 2026-07-09
Add SLURM v2 input filtering

Koenvh1:
As mentioned in #1018, this PR adds SLURM v2 input. Routinator already supports SLURM v2 as output. Relies on https://github.com/NLnetLabs/rpki-rs/pull/380

### DIFF
--- a/src/payload/validation.rs
+++ b/src/payload/validation.rs
@@ -752,7 +752,11 @@ impl<'a> SnapshotBuilder<'a> {
     fn process_aspa(&mut self, aspa: PubAspa, metrics: &mut AllVrpMetrics) {
         metrics.update(|m| m.aspas.valid += 1);
 
-        // SLURM filtering goes here ...
+        // Is the ASPA to be filtered locally?
+        if self.exceptions.drop_aspa(aspa.customer) {
+            metrics.update(|m| m.aspas.locally_filtered += 1);
+            return
+        }
 
         match self.aspas.entry(aspa.customer) {
             hash_map::Entry::Vacant(entry) => {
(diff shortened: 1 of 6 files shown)


src/slurm.rs
```diff
@@ -161,6 +190,12 @@ impl LocalExceptions {
         })
     }
 
+    pub fn drop_aspa(&self, customer: Asn) -> bool {
+        self.aspa_filters.iter().any(|filter| {
+            filter.drop_aspa(&Aspa::new(customer, ProviderAsns::empty()))
```
Koenvh1: I'm tempted to rewrite this in rpki-rs to just take a customer ASN. It extracts it from the Aspa object anyway and ignores the ProviderAsns

src/payload/validation.rs
```diff
@@ -825,7 +829,26 @@ impl<'a> SnapshotBuilder<'a> {
             }
         }
 
-        // XXX ASPA assertions.
+        for (aspa, info) in self.exceptions.aspa_assertions() {
+            match self.aspas.entry(aspa.customer) {
+                hash_map::Entry::Vacant(entry) => {
+                    entry.insert((
+                        SmallAsnSet::from_iter(aspa.providers.iter()),
```
Koenvh1, line 836: Maybe we want a ProviderAsns into SmallAsnSet implementation?

src/slurm.rs
partim: That seems like the right thing to do – maybe add a `drop_customer` method there.

src/payload/validation.rs
partim: Probably?

PR #1107, opened 2026-07-23
Use /var/cache instead of /var/lib for storage

cvengler:
This commit performs a tree-wide edit that changes all occurrences of "/var/lib" to "/var/cache".

The reason for that is, that according to the Linux File System Hierarchy standard, "/var/cache" is intended for cached data from applications that can safely be deleted without data loss, which is the case here, given that the data can be obtained again from an upstream source.

Besides, this is a more sensible default.  Many users, including myself, perform automated backups of "/var" while excluding "/var/cache". Likewise, I do not see a reason on why I should have to include the application data of Routinator in my backups, which probably justifies using "/var/cache" for this data instead.

I hope that this change is not too late and/or too breaking but I feel that this is the more sensible choice to do.


partim:
I could swear I read the FHS to figure out where to put the data, but you are right. (I think initially we also had the TALs in this directory, so maybe it did make sense at the time.)

We have to figure out a way to migrate to the new place when upgrading the Debian and RPM packages we are distributing.

cvengler:
Sorry if that PR was a bit too hasty, maybe we should track it in a separate ticket?

partim:
No worries, the PR is helpful. We’ll do some research internally to figure things out and then either extend this PR or add another one. It’ll be a couple weeks, though, due to vacation and such. For now, I suggest we just keep the PR open.

Koenvh1:
I agree this is probably a good idea. I _think_ the only thing needed is to extend the systemd ReadWritePaths with both /var/lib/routinator/ and /var/cache/routinator/, it shouldn't overwrite the config after an update (nor do I want to move files of an existing installation in case people rely on things existing in /var/lib). Additionally, I'd have to check whether this holds:

>  The application must always be able to recover from manual deletion of these files (generally because of a disk space shortage).

> Any application should take into account that the file stored here can disappear any time, and be ready to recompute its contents (with some time penalty).

I don't think that should be a problem, but there might be certain Routinator stages where removing files halfway through might be problematic.

partim:
That’s a good point: Deleting stuff during a validation run will indeed cause issues.

cvengler:
@Koenvh1 May I ask where the second quote is from? I cannot find it on: https://specifications.freedesktop.org/fhs/latest/varCache.html

The only thing I can find there regarding this is:
> The data must remain valid between invocations of the application and rebooting the system.

Which to me means, that it is a bad idea to delete any /var/cache data while a process is running?

Koenvh1:
https://web.archive.org/web/20130511220135/http://www.lindevdoc.org/wiki//var/cache

Not sure how correct that is though.

job:
The Debian maintainer of the rpki-client package  (who I consider an authority on all things Debian) decided to put the cache in `/var/lib/rpki-client/cache` ... because the cache state is supposed to be durable and persist from run to run, and apparently there is no guarantee of persistence in `/var/cache` (in the Linux FHS worldview?).

On OpenBSD we use `/var/cache/rpki-client` for the cached state and `/var/db/rpki-client` for the output artifacts (such as JSON, OpenMetrics, & CCR).

The maintainer for Fedora/EPEL/Centos/Redhat universe decided on `/var/cache/rpki-client` for the cached state and `/var/lib/rpki-client` for output artifacts.

On Debian the TALs & constraints are stored in `/etc/tals/`, on Fedora/Redhat/Oracle in `/etc/pki/tals/`, and on OpenBSD in `/etc/rpki/`.

I have no specific recommendation what's best for routinator but I figured I could share some additional data points decided on by other people. I do have a feeling that if we'd ask 5 more people we'd get 5 more answers ;-)

PR #1109, opened 2026-08-03
Rewrite output formatting.

partim:
This PR rewrites output formatting to using async functions rather than a complex state machine.

In order to be able to use the same async function for both the HTTP streaming case and writing to files directly, a little bit of trickery is involved, but I think not having to write the functions twice is more important.

This PR is still work in progress. For how the new functions look, have a look at the new ´crate::output::write_cvs` function.


src/http/payload.rs
```diff
@@ -74,9 +73,11 @@ impl State {
             Ok(res.empty())
         }
         else {
-            Ok(res.stream(
-                stream::iter(output.stream(snapshot, metrics, format))
-            ))
+            let (writer, response) = res.stream_frames(0xFFFF);
```
Koenvh1: What's the reason for the frame size `0xFFFF`?

src/http/payload.rs
```diff
@@ -74,9 +73,11 @@ impl State {
...
+            tokio::spawn(async move {
```
Koenvh1, line 77: There is probably a very good reason, but why is this inside `tokio::spawn` and not just awaited?

src/http/payload.rs
partim: No reason, I just though 64k is a good number. I might be wrong.

src/http/payload.rs
partim: You can’t really await at this point – the response needs to take each generated frame but it is returned and that taking happens somewhere else.

However, the function itself doesn’t need to be async.

partim:
We should switch metrics and status writing to this new approach as well before merging.

